PR #113, opened 2026-05-12
Fix onboarding window vanishing when 'Open Settings' is clicked

spencerogden:
## Summary
- Drop the `window.orderBack(nil)` call from `OnboardingWindowController.yieldFocusToSystemSettings()`.
- De-floating the window via `window.level = .normal` is enough to let System Settings render on top — sending the window to back was the cause of the crash-like symptom.

## Why
On macOS 26 (Tahoe), clicking the "Open Settings" button on any onboarding permission step (Microphone, Accessibility, Input Monitoring) ghost-stated the entire app:
- The onboarding window vanished
- The dock icon disappeared
- Cmd+Tab no longer included MuesliDev
- No menu bar icon yet (onboarding hadn't finished, so the status bar item wasn't registered)
- `pgrep -lf MuesliDev` confirmed the process was still alive — but with **zero UI surface**, the only recovery was `kill <pid>` and relaunch

Root cause: the previous `level = .normal` + `orderBack(nil)` combination on a previously-`.floating` window in an `.accessory`-policy app makes AppKit revoke the temporary dock representation. The window is technically still ordered (just at the back), but with no dock and no Cmd+Tab entry the user has no way to find or recover it.

## How I hit this
Verifying another PR locally with `MUESLI_SKIP_SIGN=1 ./scripts/dev-test.sh` triggered fresh onboarding (the unsigned binary's cdhash didn't match the previously-granted TCC entry). Clicking "Open Settings" on the Accessibility step → app vanished → no recovery path. `sample <pid>` showed the main thread idle in the AppKit run loop, no crash dump in `~/Library/Logs/DiagnosticReports/`, just a healthy zombie.

## Fix
```diff
 func yieldFocusToSystemSettings() {
     guard let window else { return }
     window.level = .normal
-    window.orderBack(nil)
 }
```

System Settings still gets focus naturally when launched via `NSWorkspace.shared.open(URL("x-apple.systempreferences:..."))`. The onboarding window stays at standard window level — visible behind System Settings, clickable to bring forward, and crucially **still represented in the dock and Cmd+Tab**.

## Validation
- Reproduced the bug on macOS 26.4.1 with the unfixed code.
- Applied the fix, rebuilt with `MUESLI_SKIP_SIGN=1 ./scripts/dev-test.sh`.
- Walked through onboarding fresh (after `tccutil reset All com.muesli.dev`):
  - Clicked "Open Settings" on the Accessibility step
  - System Settings opened in front, onboarding window stayed visible behind
  - Granted Accessibility, polling detected within ~1s, "Continue" appeared
  - App restarted normally and onboarding finished

No new tests — this is UI window-management behavior that the existing test suite doesn't cover (no SwiftUI/AppKit window tests in the repo). The change is small enough to verify by code inspection + the manual repro above.

---

<a href="https://app.devin.ai/review/muesli-hq/muesli/pull/113" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an onboarding window issue where the window could be pushed behind other windows.
  * Improved the handoff to system settings so the onboarding screen stays accessible and the app remains recoverable.


pHequals7:
Thanks for the onboarding fix. This PR is otherwise ready, but the DCO check is required for all contributions.

Could you please rebase onto the current `main`, add your own sign-off to the commit, and force-push the updated branch? For example:

```bash
git rebase origin/main
git commit --amend --signoff --no-edit
git push --force-with-lease
```

The sign-off needs to come from the original contributor, so we cannot add it on your behalf. Once the DCO check passes, we can merge this.

PR #132, opened 2026-05-14
Fix meeting auto-stop and background transcription blocking

rogersterling:
## Summary
- Track meeting activity separately from prompt suppression so active recordings can auto-stop when the meeting source disappears.
- Carry the auto-stop source through detected, scheduled, auto-recorded, and join-and-record meeting starts.
- Allow dictation and computer-use commands while a stopped meeting is transcribing in the background, while still blocking them during active meeting audio capture.

## Testing
- Manually verified with `MuesliDev`: leaving/closing a Google Meet auto-stops recording, and dictation works while meeting transcription continues.
- `env -u OPENAI_API_KEY -u OPENROUTER_API_KEY swift test --package-path native/MuesliNative --scratch-path "$HOME/Library/Caches/muesli-spm/test"`

### DIFF
--- a/native/MuesliNative/Sources/MuesliNativeApp/MeetingAutoStopPolicy.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/MeetingAutoStopPolicy.swift
@@ -1,0 +1,70 @@
+import Foundation
+
+struct MeetingAutoStopSource: Equatable {
+    let candidateID: String?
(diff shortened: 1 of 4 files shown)


rogersterling:
Closing this combined PR in favor of two smaller PRs: #133 for meeting auto-stop and #134 for allowing dictation while meeting transcription finishes.

PR #135, opened 2026-05-14
[codex] Exclude dictation mic from meeting detection

rogersterling:
## Summary

- Filter Muesli-owned mic/audio attribution before resolving meeting detection candidates.
- Keep external meeting-app media signals eligible, including cases where Muesli is also using input.
- Add regression coverage for dictation during an active calendar window so it does not create a meeting candidate.

## Root Cause

Meeting detection treated the global mic device as meeting evidence even when Muesli itself was the only attributed mic owner. During a current or nearby calendar event, using dictation could therefore satisfy `calendar + mic` and surface a false meeting detection prompt.

## Validation

- `swift test --package-path native/MuesliNative --scratch-path "$HOME/Library/Caches/muesli-spm/test" --filter MeetingMediaSignalFilter`
- `env -u OPENAI_API_KEY -u OPENROUTER_API_KEY swift test --package-path native/MuesliNative --scratch-path "$HOME/Library/Caches/muesli-spm/test"`

### DIFF
--- a/native/MuesliNative/Sources/MuesliNativeApp/MeetingMonitor.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/MeetingMonitor.swift
@@ -283,6 +283,68 @@ private struct MeetingDetectionEvaluationContext {
     )
 }
 
+struct MeetingMediaSignals: Equatable {
+    let micActive: Bool
+    let cameraActive: Bool
+    let audioInputProcesses: [AudioProcessActivity]
+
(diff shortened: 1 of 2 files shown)


pHequals7:
Closed as superseded by #157, which merged the accepted meeting lifecycle/UX work into the maintainer-owned compound PR.\n\nThanks for the contribution. The relevant commits/authorship were preserved in #157 where applicable.

PR #149, opened 2026-05-15
Compact single-line layout for Dictionary view

Gui11aum3:
## Summary

The Dictionary settings page currently uses multi-row cards — each entry takes 3 rows (word fields, slider, action buttons), making it hard to scan when you have many custom words.

This PR collapses each entry into a single horizontal row:

**Before:** Word + replacement fields → full-width threshold slider → Delete/Save buttons (3 rows per entry)

**After:** `[Match] → [Replace] [85% ▾] [✓] [🗑]` (1 row per entry)

Changes:
- Replace the threshold slider with a compact dropdown menu (5% increments: 70%–95%)
- Inline all controls in a single `HStack`: text fields, arrow, threshold picker, save checkmark, delete trash icon
- Add column headers (Match / Replace / Threshold) for scannability
- Save checkmark is dimmed until there are unsaved changes, then highlights in accent color
- Same compact layout for the "Add new" inline row (checkmark to confirm, X to cancel)

## Motivation

With 10+ dictionary entries (common for non-English names and domain terms), the current layout requires excessive scrolling. The threshold slider gives false precision — most users pick round values. A dropdown with 5% steps covers the useful range and takes a fraction of the space.

## Test plan

- [ ] Open Dictionary tab — verify column headers and compact rows render correctly
- [ ] Edit a word's match/replacement — verify checkmark highlights, click saves
- [ ] Change threshold via dropdown — verify it updates and shows as unsaved
- [ ] Delete an entry — verify trash icon removes it
- [ ] Click "+ Add new" — verify inline add row appears with same compact layout
- [ ] Cancel adding — verify X button clears and hides the row
- [ ] Verify with 0, 1, and 10+ entries


Gui11aum3:
@pHequals7 : Do you need more details to merge? 
There's a bunch of open PR - if we can help to make this into a new release this would be awesome.

PR #133, opened 2026-05-14
Auto-stop meeting recording when source disappears

rogersterling:
## Summary
- Track meeting activity separately from prompt visibility so active recordings can keep observing the underlying meeting source while prompts are suppressed.
- Arm recordings with an auto-stop source from detected meetings, calendar notifications, auto-recording, and join-and-record flows.
- Stop the active meeting recording after the source disappears for a short grace period.

## Testing
- Manually verified with `MuesliDev`: leaving/closing a Google Meet auto-stops recording.
- `env -u OPENAI_API_KEY -u OPENROUTER_API_KEY swift test --package-path native/MuesliNative --scratch-path "$HOME/Library/Caches/muesli-spm/test"`

PR #157, opened 2026-05-19
Improve meeting recording lifecycle reliability

pHequals7:
## Summary

This compound PR brings together the meeting recording lifecycle reliability work and the follow-up UX cleanup from local testing.

- Auto-stop meeting recordings when the observed meeting source disappears after the grace period.
- Harden meeting detection against Muesli-owned dictation/camera signals so self activity does not keep a meeting alive or create false positives.
- Add a preparing state for meeting transcription startup so the floating indicator and meeting detail controls show clear feedback before the mic/transcription state is live.
- Improve meeting notes rendering, constrain the notes content layout, keep the sidebar behavior native/balanced, and prevent non-overflowing titles from marquee scrolling.
- Improve generated meeting titles by sampling opening, middle, and closing transcript excerpts instead of over-weighting the beginning.

## Linked PRs

Closes #133.
Closes #135.
Closes #139.

Note on #139: the green menu-bar activation indicator was tested and intentionally reverted in this compound branch because it duplicated existing recording affordances and felt visually excessive. The original #139 commit remains in branch history with a revert so contribution history is preserved while final behavior omits the green indicator.

## Contributor Credit

This branch preserves the original commit authors from the linked PR work:

- Roger Sterling for the meeting auto-stop lifecycle and self-owned signal hardening work.
- walton_UBER for the menu-bar indicator experiment that was evaluated and then reverted based on UX review.

If this PR is squash-merged, please keep the generated co-author trailers from the PR commits so contributor attribution remains attached to the merged commit.

## Validation

- `swift test --package-path native/MuesliNative --scratch-path "$HOME/Library/Caches/muesli-spm/test-meeting-lifecycle" --filter 'MeetingSummaryClient|MeetingAutoStopPolicyTests|MeetingMediaSignalFilterTests|MeetingCandidateResolverTests|MeetingPromptStateMachineTests'`
- 79 tests passed across 5 suites.
- Rebuilt and launched MuesliDev locally during testing with `MUESLI_SWIFTPM_SCRATCH_PATH="$HOME/Library/Caches/muesli-spm/dev-meeting-lifecycle" ./scripts/dev-test.sh`.

## Out of Scope

A separate debug pass found that system-audio capture can produce a silent system WAV during far-end-only speaker playback, which starves system transcription, diarization, and AEC reference. That investigation is intentionally kept out of this lifecycle PR and should be handled in a separate branch.

### DIFF
--- a/native/MuesliNative/Sources/MuesliCore/DictationStore.swift
+++ b/native/MuesliNative/Sources/MuesliCore/DictationStore.swift
@@ -727,6 +727,28 @@ public final class DictationStore {
         }
     }
 
+    public func updateMeetingTranscript(id: Int64, rawTranscript: String) throws {
+        let db = try openDatabase()
+        defer { sqlite3_close(db) }
+        let manualNotes = try manualNotesForMeeting(id: id, db: db)
+        let wordCount = Self.countWords(in: rawTranscript) + Self.countWords(in: manualNotes)
+        let sql = "UPDATE meetings SET raw_transcript = ?, word_count = ? WHERE id = ?"
+        var statement: OpaquePointer?
+        guard sqlite3_prepare_v2(db, sql, -1, &statement, nil) == SQLITE_OK else {
+            throw lastError(db)
+        }
+        defer { sqlite3_finalize(statement) }
+        sqlite3_bind_text(statement, 1, (rawTranscript as NSString).utf8String, -1, nil)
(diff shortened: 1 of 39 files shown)


dbwecon:
@pHequals7 what do you think of having the green circle while recording in the menu bar be a togglable setting? i am fine having it off by default if you think it's distracting, but i do have a preference for it. if you think it adds unnecessary complexity to the code, i am open to discussing design.

PR #171, opened 2026-05-23
Testing iOS app integration with the macOS app

pHequals7:
small step for muesli
big step for private on device dictation


Adithya2907:
Oh wow, this is very nice! Great to know that iOS application is being worked on! What's the plan for testing and rolling this out? Will we be using Testflight?

pHequals7:
> Oh wow, this is very nice! Great to know that iOS application is being worked on! What's the plan for testing and rolling this out? Will we be using Testflight?

yes checkout Muesli-HQ/muesli-ios repo

PR #182, opened 2026-05-27
feat: live transcript readout, in-meeting chat, and AEC memory fixes

### LINKED ISSUE #165: Add live chat with the transcript
This is such a great app. One feature that would make it the goat is the ability to start a meeting and chat with the transcript/diarized chunks. 

Google Meet, Copilot, Zoom AI companion, etc all have this feature but they obviously rely on the cloud provider. There is an open source project call MacParakeet that has this feature, but it does not have the quality UI, diarization, and extensibility of Muesli.

Is this something we can add?

kennyparsons:
## Summary

- **Live transcript tab**: closed-caption-style scrolling view during meetings, labelled "You" (mic) and "Others" (system audio), auto-scrolls to latest chunk
- **In-meeting chat**: query the live transcript with multi-turn LLM conversation while recording; switches to `raw_transcript` post-meeting with full diarization labels
- **Add `scripts/monitor-memory.sh`**: automated process memory health check — samples footprint, Float array sizes, AG node count, and tracking dict count on a configurable interval, with a summary verdict (STABLE / SLOW GROWTH / LEAKING)

## Test plan

- `swift test --package-path native/MuesliNative` — 917 tests, all pass except 2 pre-existing flaky `HotkeyMonitor` timing tests on main
- `./scripts/monitor-memory.sh 180 15` — run during a meeting to verify footprint stays flat and largest Float array stays ≤ ~150,000 items
- Start a meeting, speak → switch to **Live** tab → verify chunks appear with timestamps and speaker labels, auto-scroll works
- Switch to **Chat** tab → ask a question → verify LLM responds using transcript content, follow-up preserves history
- Stop meeting → verify live transcript clears, Notes/Transcript tabs resume normally


kennyparsons:
@pHequals7 Take a look and give it a test.

kennyparsons:
I think the biggest feedback we should look for is memory consumption. I think I've prevented any memory leaks or ballooning but the more people to use it, the better. Can this potentially be released as a beta?

pHequals7:
I think memory consumption in my local tests seems to be fine (~155-170 MB) - would be iterating on the visual representation but overall this was a great PR - much appreciated

kennyparsons:
I'm sure you saw it, but I consolidated the LLM networking into `LLMBackendClient.swift`. This establishes a single, shared HTTP transport layer for all backends (OpenAI, OpenRouter, Ollama, etc.), making the LLM capabilities much easier to maintain and reuse across the application. Future updates to these providers will be immediately available to the rest of the application, without needing to update the individual call sites.

kennyparsons:
Also, I think it would be worth moving llm backends to a separate config page, rather than being embeded into the meeting settings. Reason: it can/might be used in more than just meetings. 

For example, dictations could use a backend to clean up transcripts. This is a feature of TypeWhisper, a fantastic mac app that I currently use for dictations only. 

Other apps, like RayCast, have a similar approach, specifying a default backend, and then allowing per-function overrides. For example, a dictation might use ollama, to save on costs, while a meeting transcript might use a different local/online model for better accuracy/faster processing.

pHequals7:
> Also, I think it would be worth moving llm backends to a separate config page, rather than being embeded into the meeting settings. Reason: it can/might be used in more than just meetings.
> 
> For example, dictations could use a backend to clean up transcripts. This is a feature of TypeWhisper, a fantastic mac app that I currently use for dictations only.
> 
> Other apps, like RayCast, have a similar approach, specifying a default backend, and then allowing per-function overrides. For example, a dictation might use ollama, to save on costs, while a meeting transcript might use a different local/online model for better accuracy/faster processing.

Fair push - the only reason haven't actively worked on consolidating the LLM backend was the concept of privacy gets a bit lost but making it user opt-in makes complete sense

mfoll:
This PR looks very useful. The live transcript/chunk view is exactly the missing piece for long meetings.

One question about failure recovery: are live transcript chunks persisted anywhere before the meeting is stopped normally?

From a quick read, it looks like completed chunks are collected in memory (`MeetingChunkCollector`) and appended to `appState.liveMeetingTranscript`, then the durable `rawTranscript` is only written when `completeLiveMeeting(...)` runs after a clean stop. On app relaunch, stale `recording` / `processing` meetings seem to be marked as `failed`.

For long meetings, it would be very helpful if live chunks were durably saved as they arrive, so a crash, forced quit, sleep/wake problem, or app relaunch can recover a partial transcript instead of losing everything since the last clean stop.

Would you consider adding one of these as part of this PR or a follow-up?

- persist each live chunk to the meeting row or a `meeting_live_chunks` table as soon as it is transcribed;
- keep chunk audio under Application Support with the `meeting_id`, rather than only anonymous temp files;
- on relaunch, offer to recover stale `recording` / `processing` meetings from saved chunks/audio;
- if full recovery is out of scope, at least preserve the partial live transcript and show it on the failed meeting.

This would make the live transcript feature much safer for 1-2 hour meetings where losing the in-progress transcript would be painful.

kennyparsons:
> are live transcript chunks persisted anywhere before the meeting is stopped normally?

Not now. Live transcript is more like a "closed caption" with the added benefit of being able to chat with the transcript. Persisting each chunk would require more work and I didn't see a need for it when I made the PR. After it merges and is in main/prod, I'm sure we can add it, or you can have a go at it too.

pHequals7:
> This PR looks very useful. The live transcript/chunk view is exactly the missing piece for long meetings.
> 
> One question about failure recovery: are live transcript chunks persisted anywhere before the meeting is stopped normally?
> 
> From a quick read, it looks like completed chunks are collected in memory (`MeetingChunkCollector`) and appended to `appState.liveMeetingTranscript`, then the durable `rawTranscript` is only written when `completeLiveMeeting(...)` runs after a clean stop. On app relaunch, stale `recording` / `processing` meetings seem to be marked as `failed`.
> 
> For long meetings, it would be very helpful if live chunks were durably saved as they arrive, so a crash, forced quit, sleep/wake problem, or app relaunch can recover a partial transcript instead of losing everything since the last clean stop.
> 
> Would you consider adding one of these as part of this PR or a follow-up?
> 
> * persist each live chunk to the meeting row or a `meeting_live_chunks` table as soon as it is transcribed;
> * keep chunk audio under Application Support with the `meeting_id`, rather than only anonymous temp files;
> * on relaunch, offer to recover stale `recording` / `processing` meetings from saved chunks/audio;
> * if full recovery is out of scope, at least preserve the partial live transcript and show it on the failed meeting.
> 
> This would make the live transcript feature much safer for 1-2 hour meetings where losing the in-progress transcript would be painful.

I think it is worth preserving the raw transcript without any diarization as only an edge/failure case (sudden app failure/crash) - I am also in the process of incorporating parakeet-flash that supports streaming transcription so this method will also undergo few changes once the LLM backend is consolidated and each feature of muesli can have custom LLM backends

pHequals7:
@claude review

pHequals7:
Thanks Kenny, we tested this locally and I think the live transcript + checkpoint fallback + AEC memory fixes are in a good place now. One scope note before merging: I am intentionally not bringing in the meeting chat part as part of this PR yet. This PR is already fairly large in terms of audio/session/transcript recovery changes, so I’d rather land the live transcript and memory fixes cleanly here and handle chat as a separate, more focused PR once the LLM/backend flow is designed properly.

PR #186, opened 2026-05-28
Fix remaining org transfer URLs

pHequals7:
## Summary
- Update the app fallback download URL to the new GitHub Pages site
- Update the release checklist appcast verification URL
- Update README Star History links to the org-owned repo

## Validation
- git diff --check
- ./scripts/verify_update_flow.sh --skip-dmg

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -21,8 +21,10 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      pull-requests: read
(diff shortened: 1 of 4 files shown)


pHequals7:
Note on the latest Claude run: this PR now modifies the Claude review workflow itself to enable visible reporting: use_sticky_comment, track_progress, and display_report. Claude app validation rejects PR runs where the workflow file differs from the default branch, so the failure on this PR is expected. After this merges to main, the next ordinary PR should exercise the new reporting settings.

PR #208, opened 2026-06-11
Replace TelemetryDeck analytics with a local-only event log

lorenzojb:
## What this does

Removes the TelemetryDeck SDK entirely and replaces it with a local JSONL event log. Usage events are written to `events.jsonl` in the app support directory and never leave the machine. After this change the app makes no analytics network requests at all.

## Why

TelemetryDeck was initialized unconditionally on every launch with no opt-out, sending feature-usage and permission-grant events to nom.telemetrydeck.com. The event taxonomy itself is useful for local debugging, so the instrumentation points are kept — only the destination changes.

## What changed

- **Removed** the TelemetryDeck/SwiftSDK package dependency, its initialization in `AppDelegate`, and all nine `TelemetryDeck.signal` call sites.
- **Added** `LocalTelemetry` / `EventLogWriter`: appends one JSON line per event (`ts`, `event`, `app_version`, optional `params`) on a background serial queue. Files are opened with `O_CREAT|O_APPEND` and mode 0600; permissions are re-tightened on every append in case the file pre-existed with a looser mode. The log is capped at 5 MB with a single rotation slot (`events.jsonl.1`), so disk use is bounded and the newest events are always retained.
- **No identifiers**: unlike TelemetryDeck there is no install ID, session ID, or device metadata — just timestamp, event name, app version, and the existing per-event params.
- **Opt-out**: new `enable_event_log` config key (default true). Logging stays disabled until `MuesliController` applies the loaded config, and the flag is re-applied on every `updateConfig`, so toggling it takes effect immediately without a restart.
- **Lifecycle**: `app.launched` now fires after controller startup (so the enable flag is respected); pending writes are flushed in `applicationWillTerminate` so events aren't dropped on quit.
- **Privacy policy** (`docs/privacy.html`) updated to describe the local log instead of TelemetryDeck.

## How it was verified

- Full test suite passes: 991 tests in 112 suites (7 new tests covering line format, enable/disable toggling, 0600 permissions on create and on pre-existing world-readable files, rotation at the size cap, and directory creation).
- Grep confirms no remaining TelemetryDeck references in sources, scripts, or CI.

## After merge

`events.jsonl` starts accumulating locally for users on the next update; anyone who wants it off sets `enable_event_log: false` in config.json. A possible follow-up is a `muesli-cli events` reader, which would argue for moving `EventLogWriter` into MuesliCore.

---
<a href="https://app.blacksmith.sh/Muesli-HQ/codesmith/muesli/pr/208"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783762277&installation_id=136549638&pr_number=208&repository=Muesli-HQ%2Fmuesli&return_to=https%3A%2F%2Fgithub.com%2FMuesli-HQ%2Fmuesli%2Fpull%2F208&signature=305c34a799f1541a475f1152b8fdc8c97a21c8230e122781215d81df3f38ce5e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/docs/privacy.html
+++ b/docs/privacy.html
@@ -162,11 +162,11 @@
     <li><strong>Calendar</strong> &mdash; to read upcoming events from your local macOS calendar (EventKit)</li>
   </ul>
 
-  <h2>Anonymous Usage Analytics</h2>
-  <p>Muesli uses <a href="https://telemetrydeck.com">TelemetryDeck</a> to collect minimal, anonymized usage signals (e.g., which features are used, app launch counts). TelemetryDeck is privacy-first by design &mdash; it does not collect IP addresses, device identifiers, or any information that could identify you personally. No audio, transcripts, calendar data, or personal content is ever included in analytics. You can learn more about TelemetryDeck's privacy practices on their <a href="https://telemetrydeck.com/privacy">privacy page</a>.</p>
+  <h2>Local Usage Log</h2>
+  <p>Muesli does not use any analytics service. Minimal usage events (e.g., which features are used, app launch counts) are recorded only to a local file on your Mac (<code>events.jsonl</code> in the application's support directory) and never leave your device. No audio, transcripts, calendar data, or personal content is included. You can disable this log entirely by setting <code>enable_event_log</code> to <code>false</code> in the application's <code>config.json</code>.</p>
 
   <h2>No Personal Data Collection</h2>
-  <p><strong>Muesli does not collect, transmit, or store any personal data on external servers.</strong> Apart from the anonymous usage signals described above, no data leaves your device. The application is fully functional without an internet connection (except for optional cloud summarization and calendar sync).</p>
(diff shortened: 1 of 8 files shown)


lorenzojb:
Opened against the wrong repo, closing.

Issue #172, opened 2026-05-23
Feature request : Music pauses while dictating [toggle option]


xhitijc-meesho:
Just like wispr flow, it should have an option to pause the muesli when dictating and then again resume when dictation is done over so that the user can focus on dictating.

pHequals7:
hey this should be coming in the next update - but thanks for flagging!

pHequals7:
Hi this has been incorporated in v0.6.8 - currently it only supports pausing since resuming API is not very reliable (happy to see a PR if someone can figure that out)

<img width="801" height="127" alt="Image" src="https://github.com/user-attachments/assets/9b5f3f69-abfc-4b83-8261-ff830f492519" />

PR #226, opened 2026-06-19
Fix pause-media-during-dictation un-pausing already-paused media (#225)

### LINKED ISSUE #225: Pause media during dictation: already-paused media gets un-paused on press and re-paused on release
## Description

With **"Pause media during dictation"** enabled (Settings → Advanced), starting a dictation while media is **already paused** incorrectly *starts* playback on hotkey press, then *re-pauses* on release. Instead of being a no-op (nothing is playing, so nothing should be paused/resumed), the feature inverts the media state for the duration of the hold.

The intended behavior is: if audio is playing → pause on press, resume on release; if nothing is playing → do nothing.

## Reproduction

1. Enable **Settings → Advanced → "Pause media during dictation"**
2. Open a video/track in Safari, Chrome, QuickTime, etc. and **pause it**
3. Hold the dictation hotkey, speak, release
4. **Observed:** the paused media starts playing the instant you press the hotkey, and re-pauses when you release
5. **Expected:** the already-paused media stays paused throughout — the feature should be a no-op when nothing is playing

## Root cause

The gate that decides whether to send a play/pause toggle relies on `outputActivityStatus()`, which reads `kAudioProcessPropertyIsRunningOutput` per process (`MediaPlaybackController.swift`):

```swift
// beginDictationMediaPause
guard client.outputActivityStatus() == .active else { return }
client.sendMediaPlayPauseToggle()
```

`IsRunningOutput` means **"the app's audio output pipeline is running"**, not **"audio is currently audible."** Browsers and most video players keep their audio engine/IO alive while a video is *paused* — they don't tear it down. So a paused video still reports `IsRunningOutput == true`, and `outputActivityStatus()` returns `.active`.

Consequently:

1. **Press** → status reads `.active` (paused, but IO alive) → guard passes → `sendMediaPlayPauseToggle()` → **paused media starts playing**, `pausedForSession = true`.
2. **Release** → `restoreDictationMediaPause()` runs inside the ducking-restore completion (after the ~0.5s codec-stabilization delay in `AudioDuckingController`). Its guard is `outputActivityStatus() != .active`; during that transition window output reads as not-active → guard passes → another toggle → **media re-pauses**.

The play/pause key is a blind global toggle, so once the activity check guesses wrong on press, the controller has no reliable way to undo it. The existing code comment already acknowledges this: *"macOS exposes a reliable public media key toggle, not a global 'resume only what I paused' API."*

## Core flaw

`IsRunningOutput` cannot distinguish **"actively producing sound"** from **"paused but holding an idle audio engine."** Any state-detection built on it will misfire for the very common "app open, media paused" case.

## Suggested fix

Use a playback-state signal that reflects whether audio is *actually* playing before sending the toggle, e.g.:

- Read the system now-playing state (via `MediaRemote`'s now-playing info / playback-rate) and only toggle when something reports it is actively playing, or
- Sample the output device's actual signal activity rather than relying on the per-process `IsRunningOutput` flag.

At minimum, `beginDictationMediaPause` should be conservative: if it cannot positively confirm audio is playing, it should **not** toggle (no-op) rather than risk un-pausing already-paused media.

## Related

Implements the toggle requested in #172.

## Environment

- Muesli v0.7.0
- macOS 15.6 (arm64, Apple Silicon)
- `pause_media_during_dictation = true`

pHequals7:
## Summary

Fixes #225.

With **Settings → Advanced → Pause media during dictation** enabled, starting a dictation while media was **already paused** incorrectly *started* playback on hotkey press and *re-paused* on release, instead of being a no-op.

## Root cause

The pause decision used `outputActivityStatus()`, which reads `kAudioProcessPropertyIsRunningOutput` per process. That flag means **"the app's audio output pipeline is running"**, not **"audio is currently audible."** Browsers and most video players keep their audio engine/IO alive while a video is *paused*, so paused media reported `.active`. The play/pause media key is a blind global toggle, so the press started the paused media and the release re-paused it.

## Fix

Gate the toggle on the system now-playing playback state instead of `IsRunningOutput`, using MediaRemote's `MRMediaRemoteGetNowPlayingApplicationIsPlaying` (loaded via `dlsym`, like the existing LocalVQEBridge private-framework pattern), which reflects whether audio is *actually* playing.

`MediaPlaybackClient` now exposes `nowPlayingPlaybackState() -> MediaPlaybackState { playing | notPlaying | unknown }`.

- **`beginDictationMediaPause`**: only send the toggle when now-playing positively reports `.playing`. Paused/stopped (`notPlaying`) or `unknown` → **no-op**, so already-paused media stays paused (the conservative behavior the issue asks for: do not toggle unless playback is positively confirmed).
- **`restoreDictationMediaPause`**: resume what Muesli paused, but skip the resume only when something is actively playing again (the user resumed/started playback) so we don't pause it. `unknown` still restores, matching the prior "restore what we paused over stranding playback" intent. The unreliable `IsRunningOutput` re-check is removed, because a paused video keeps reporting its engine as running and would block the resume.

The dead CoreAudio process-scan code (`processObjectIDs`/`pidProperty`/`boolProperty`/`getUInt32`) is removed from `SystemMediaPlaybackClient`; it was only used for the now-replaced activity check. The ducking controller's separate `AudioOutputActivityStatus`/`outputActivityStatus()` is untouched (ducking a paused/silent output is harmless and out of scope for this issue).

## Tests

`MediaPlaybackControllerTests` updated for the new protocol and semantics:
- New regression test: **already-paused media (`notPlaying`) is left alone throughout** — 0 toggles on press and release (the core bug).
- New test: **unknown playback state is a no-op on press** — 0 toggles.
- Existing cases remapped: playing→pause+restore (2 toggles); restore skipped when user resumed (`playing` on release → 1 toggle); restore resumes on `unknown` (2 toggles); duplicate begin pauses once; disabled/headphone skipped.

`DictationAudioSessionManagerTests` use `MediaPlaybackManaging` (unchanged), so they are unaffected.

## Verification status

- ⚠️ **Not built/run in this environment.** The cloud Linux sandbox has no Swift toolchain, and the package depends on macOS-only frameworks (FluidAudio, WhisperKit, Sparkle, MediaRemote via dlsym), so `swift test --package-path native/MuesliNative` cannot run here.
- Logic was traced by hand against the controller state machine; all eight controller tests pass under the new guards.
- The MediaRemote dlsym path is only exercised by the real `SystemMediaPlaybackClient`, not by the unit tests (which inject a fake client).
- **Needs CI + local MuesliDev QA** to confirm: build, the unit-test suite, and real-world behavior with Safari/Chrome/QuickTime/Music paused and playing.

## Notes for QA

- Apps that publish now-playing state (Safari, Chrome, Music, Spotify, QuickTime) will now be paused only when actually playing, and resumed on release.
- Apps that do **not** publish now-playing state will simply not be auto-paused (no-op) rather than risk un-pausing paused media — this is the intended conservative trade-off per the issue.

Related: implements the toggle requested in #172.

---
<a href="https://app.blacksmith.sh/Muesli-HQ/codesmith/muesli/pr/226"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784422556&installation_id=136549638&pr_number=226&repository=Muesli-HQ%2Fmuesli&return_to=https%3A%2F%2Fgithub.com%2FMuesli-HQ%2Fmuesli%2Fpull%2F226&signature=b3c7fa46d1350a9585ff00df7b45b225a271eda4a236dd4d500d89f023b8c8e1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---

<a href="https://app.devin.ai/review/muesli-hq/muesli/pull/226" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/native/MuesliNative/Sources/MuesliNativeApp/MediaPlaybackController.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/MediaPlaybackController.swift
@@ -1,5 +1,5 @@
 import AppKit
-import CoreAudio
+import Darwin
 import Foundation
 
 protocol MediaPlaybackManaging: AnyObject {
@@ -7,8 +7,30 @@ protocol MediaPlaybackManaging: AnyObject {
     func restoreDictationMediaPause()
 }
 
+/// Actual playback state of the system now-playing application, as opposed to
+/// `AudioOutputActivityStatus`, which only reflects whether an app's audio
+/// output pipeline is running. Browsers and most video players keep their
+/// audio engine/IO alive while a video is *paused*, so `IsRunningOutput` (and
+/// therefore `AudioOutputActivityStatus`) reports "active" for paused media
+/// and cannot be used to decide whether to send a play/pause toggle.
+enum MediaPlaybackState: Equatable, CustomStringConvertible {
+    case playing
+    case notPlaying
+    case unknown
+
+    var description: String {
+        switch self {
+        case .playing: return "playing"
+        case .notPlaying: return "not-playing"
+        case .unknown: return "unknown"
+        }
+    }
+}
+
 protocol MediaPlaybackClient {
-    func outputActivityStatus() -> AudioOutputActivityStatus
+    /// Whether the current now-playing application is actually producing audio.
+    /// `.unknown` is returned when the signal cannot be obtained.
+    func nowPlayingPlaybackState() -> MediaPlaybackState
     func sendMediaPlayPauseToggle()
 }
 
@@ -30,7 +52,13 @@ final class MediaPlaybackController: MediaPlaybackManaging {
             guard enabled else { return }
             guard !pausedForSession else { return }
             guard routeKind == .speakerLike else { return }
-            guard client.outputActivityStatus() == .active else { return }
+            // The media key is a blind global toggle: sending it to already-
+            // paused media would *start* playback. Only pause when we can
+            // positively confirm something is actually playing. IsRunningOutput
+            // cannot be used here because it stays true for paused media
+            // (browsers/video players keep their audio engine alive while
+            // paused), so an unknown state must also be a no-op.
+            guard client.nowPlayingPlaybackState() == .playing else { return }
             client.sendMediaPlayPauseToggle()
             pausedForSession = true
         }
@@ -40,12 +68,14 @@ final class MediaPlaybackController: MediaPlaybackManaging {
         queue.async { [self] in
             guard pausedForSession else { return }
             pausedForSession = false
-            // macOS exposes a reliable public media key toggle, not a global
-            // "resume only what I paused" API. Resume only when output still
-            // does not look active so we do not pause user-started playback.
-            // If activity is unknown, prefer restoring media we know Muesli
-            // paused over leaving playback stranded.
-            guard client.outputActivityStatus() != .active else { return }
+            // We only paused media we confirmed was playing, so resume it. Do
+            // not gate on IsRunningOutput here: a paused video still reports
+            // its audio engine as running, which would block the resume and
+            // strand playback. Only skip the resume when something is actively
+            // playing again (the user resumed/started playback) so we do not
+            // pause it. When state is unknown, prefer restoring media we know
+            // Muesli paused over leaving playback stranded.
+            guard client.nowPlayingPlaybackState() != .playing else { return }
             client.sendMediaPlayPauseToggle()
         }
     }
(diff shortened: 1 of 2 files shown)


pHequals7:
This is a duplicate of the base fix that already landed via #227. I opened #232 as the follow-up that keeps the #225 behavior fix while addressing the blocking/callback review concerns from #227. Recommendation: close this after #232 lands.

pHequals7:
Closing as duplicate. #227 already landed the base fix, and #232 has now merged the follow-up that addresses the review concerns while keeping #225 fixed.

PR #272, opened 2026-07-05
Bound summary prompts so long meetings stop failing

ivkiwi:
## Problem

The meeting summary prompt embeds the entire transcript (`summarizeOnce` path in `MeetingSummaryClient`). Multi-hour meetings exceed the model context window and the summary fails with an opaque provider error — on every backend. Title generation already has excerpt logic; summaries had none.

## Fix

Transcript input to the shared summary prompt builder is capped at 24,000 characters. Over-budget transcripts are middle-truncated with an explicit marker line (`[... transcript truncated: N characters omitted ...]`) so the model still sees both how the meeting started and how it ended. Under-budget transcripts are untouched, and behavior is identical across all backends since the cap lives in the shared prompt-construction point.

## Tests

Under-budget passthrough, over-budget head+tail+marker, exact boundary, omitted-count in marker. Full suite passes (1222 tests).

---
<a href="https://app.blacksmith.sh/Muesli-HQ/codesmith/muesli/pr/272"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785863388&installation_id=136549638&pr_number=272&repository=Muesli-HQ%2Fmuesli&return_to=https%3A%2F%2Fgithub.com%2FMuesli-HQ%2Fmuesli%2Fpull%2F272&signature=3e98f367e1f045e7216829669d230f362341f5405bbe0211e2739cc54c916970"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Meeting summaries now cap transcript content sent for summarization to a fixed character budget.
  * When transcripts exceed the limit, the summary prompt uses a truncation marker that indicates omitted characters while preserving the transcript’s start and end.
* **Tests**
  * Added unit tests for truncation behavior under budget, exactly at the limit, over budget, and for very small budgets where only a head prefix is returned.

### DIFF
--- a/native/MuesliNative/Sources/MuesliNativeApp/MeetingSummaryClient.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/MeetingSummaryClient.swift
@@ -359,6 +359,7 @@ enum MeetingSummaryClient {
         var marker = summaryTranscriptTruncationMarker(omittedCount: omittedCount)
         var contentBudget = maxCharacters - marker.count
 
+        // Marker length depends on the omitted-count digits, so recalculate until stable.
         while contentBudget > 0 {
             let nextOmittedCount = transcript.count - contentBudget
             guard nextOmittedCount != omittedCount else { break }
@@ -368,11 +369,18 @@ enum MeetingSummaryClient {
         }
 
         guard contentBudget > 0 else {
-            return String(marker.prefix(maxCharacters))
+            // Tiny budgets cannot fit the marker; avoid returning a misleading partial marker.
+            let boundedTranscript = String(transcript.prefix(maxCharacters))
+            logSummaryTranscriptTruncated(
+                originalCharacters: transcript.count,
+                omittedCharacters: transcript.count - boundedTranscript.count
+            )
+            return boundedTranscript
         }
 
         let openingCharacters = contentBudget / 2
         let closingCharacters = contentBudget - openingCharacters
+        logSummaryTranscriptTruncated(originalCharacters: transcript.count, omittedCharacters: omittedCount)
         return String(transcript.prefix(openingCharacters))
             + marker
             + String(transcript.suffix(closingCharacters))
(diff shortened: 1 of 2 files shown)


ivkiwi:

**CodeRabbit: fixed-point loop and tiny-budget fallback**

Thanks, this was a fair readability nit. I added a short comment explaining the fixed-point loop and changed the tiny-budget fallback so it returns a bounded transcript prefix instead of a partial marker when the marker cannot fit. I also added a small regression test for that edge case in `MeetingSummaryClientTests`.

**Greptile: missing diagnostic log on truncation**

Good catch. I added a truncation diagnostic that logs `originalChars` and `omittedChars` to both `Logger` and stderr, matching the nearby prompt-construction diagnostics. The normal truncation output remains unchanged; only the previously ambiguous tiny-budget fallback changed as noted above.

ivkiwi:
Addressed the latest review; follow-up commit pushed.

Thanks, fixed in follow-up commit `6f9a4d4b`. The omitted-count test now locates the full truncation marker range from the actual `summaryTranscriptForPrompt` output before computing retained transcript length, instead of rebuilding the marker padding in the test.

Validation: targeted `MeetingSummaryClientTests.summaryTranscriptMarkerContainsOmittedCount` passed, and the full SwiftPM suite passed with 1224 Swift Testing tests.

ivkiwi:

PR #274, opened 2026-07-05
Preserve live transcript order under async chunk completion

ivkiwi:
## Problem

Live meeting chunk transcriptions complete asynchronously, and completed segments are emitted to live consumers in **completion** order, not recording order. When chunk N+1 (short/quiet) finishes before chunk N (long/dense), the live transcript shows interleaved out-of-order text; ordering is only repaired by the sort at final drain, so the meeting looks garbled until it stops.

## Fix

Completed chunks are buffered by registration sequence and released to live callbacks only as the next contiguous sequence — recording order is preserved during the meeting, not just after it. Empty chunks still advance the sequence so silence never stalls later text. Mic and system tracks keep independent collectors, and final-drain behavior is unchanged.

## Tests

Out-of-order completion releases in registration order, empty-chunk advancement, independent mic/system tracks. Full suite passes (1222 tests).

---
<a href="https://app.blacksmith.sh/Muesli-HQ/codesmith/muesli/pr/274"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785863503&installation_id=136549638&pr_number=274&repository=Muesli-HQ%2Fmuesli&return_to=https%3A%2F%2Fgithub.com%2FMuesli-HQ%2Fmuesli%2Fpull%2F274&signature=1312c6f1b9fb8d3b6b07eb47c73d5b46c50000d90e4665f51dd990f9e2c43044"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed live transcription to release completed chunk groups in the correct sequence, even when tasks finish out of order.
  - Improved chunk retirement/drain/cancel handling to prevent stalled or missing tail chunks.
  - Made live transcript rendering deterministic by ordering entries by time (with stable tie-breaking).
  - Improved incremental transcript merging by resetting when the new text diverges from the already-parsed prefix.

- **Tests**
  - Added coverage for live-release ordering, retire/drain edge cases, per-track independence, and transcript merge decisions.

### DIFF
--- a/native/MuesliNative/Sources/MuesliNativeApp/LiveTranscriptView.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/LiveTranscriptView.swift
@@ -15,6 +15,12 @@ private struct LiveTranscriptGroup: Identifiable {
 }
 
 struct LiveTranscriptView: View {
+    enum TranscriptMergeDecision: Equatable {
+        case unchanged
+        case append(String)
+        case resetAndAppend(String)
+    }
+
     let transcript: String
     @State private var groups: [LiveTranscriptGroup] = []
     // Tracks how many characters of transcript have been parsed into groups.
@@ -64,19 +70,29 @@ struct LiveTranscriptView: View {
     }
 
     private func mergeNewContent(from newTranscript: String) {
-        if newTranscript.count < parsedLength || !newTranscript.hasPrefix(parsedTranscript) {
+        let decision = Self.transcriptMergeDecision(
+            newTranscript: newTranscript,
+            parsedTranscript: parsedTranscript,
+            parsedLength: parsedLength
+        )
+
+        let newContent: String
+        switch decision {
+        case .unchanged:
+            return
+        case .append(let suffix):
+            newContent = suffix
+        case .resetAndAppend(let transcript):
             groups = []
             parsedLength = 0
             parsedTranscript = ""
+            newContent = transcript
         }
-        guard newTranscript.count > parsedLength else {
-            return
-        }
-        let startIndex = newTranscript.index(newTranscript.startIndex, offsetBy: parsedLength)
+
         parsedLength = newTranscript.count
         parsedTranscript = newTranscript
 
-        let newMessages = TranscriptChatMessage.messages(from: String(newTranscript[startIndex...]))
+        let newMessages = TranscriptChatMessage.messages(from: newContent)
         for msg in newMessages {
             if let last = groups.last, last.speaker == msg.speaker {
                 groups[groups.count - 1] = LiveTranscriptGroup(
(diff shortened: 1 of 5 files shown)


ivkiwi:

**Greptile: Unknown retire ID can look like a blocked live chunk**

Addressed in `c207955e25900bd22d3b0a126d92595e9f62fefd`. Unknown retire IDs now return `nil`, so stale callbacks no longer look like valid known chunks with no ready output. I intentionally did not advance the sequence on an unknown ID because the collector cannot safely infer which registration slot it belongs to without risking dropped or reordered live chunks.

**CodeRabbit: Add `cancelAll()` stale-callback test**

Added the `cancelAll()` stale-callback coverage in `c207955e25900bd22d3b0a126d92595e9f62fefd`. The test now exercises a pending task that is canceled by `cancelAll()`, then verifies the late retire path returns `nil` and emits nothing. This matches the existing closed-collector behavior after drain.

**CodeRabbit: Cancel Task closures in new tests**

Cleaned this up in `c207955e25900bd22d3b0a126d92595e9f62fefd`. The affected tests now keep task handles and cancel them after registration, matching the existing test style. This is not a functional change, but it keeps the task lifecycle consistent across the collector tests.

ivkiwi:
Follow-up pushed: field testing on our side surfaced two tail-loss defects in the ordered release gate — a hung chunk task could hold all later chunks behind the contiguous gate (meeting stop awaited tasks sequentially), and the inactivity-timeout check raced progress append, dropping a chunk that completed between an empty poll and the timeout break. The collector now drains by progress past gaps, failures retire their slot, the timeout check and progress append share one lock, and stop drains fully with timeout logging. New tests are deterministic and loop-verified 50/50.

ivkiwi:
Addressed the latest review; follow-up commit pushed.

Addressed the fresh CodeRabbit items in `7fdd5d28`.

The drain-time gap test now keeps the tail chunk pending, starts `closeAndDrainSortedSegments(inactivityTimeout:)`, waits until the collector has actually entered drain, then completes the tail. That covers the missing path where a later chunk finishes during drain while an earlier slot remains stalled.

The live transcript ordering issue was valid for the visible transcript path. The existing collectors ordered each track independently, but `MuesliController` still appended live captions by callback arrival order. I fixed that at the consumer by buffering live checkpoint entries on the main actor and rendering the live transcript sorted by segment start, matching the durable checkpoint ordering. `LiveTranscriptView` now resets when the transcript update is a reordered replacement instead of an append-only suffix.

I also made the timing-sensitive drain polling test deterministic by synchronizing on the collector closing its pending set and using a wider inactivity timeout, instead of relying on one `Task.yield()` plus 50ms.

Validation: targeted `MeetingChunkCollectorTests` + `LiveTranscriptBufferTests` passed, and the full SwiftPM test suite passed with 1229 tests.

ivkiwi:
Addressed; follow-up pushed.

Done in follow-up commit `192d9fef`.

I extracted the live transcript reset/append decision into `LiveTranscriptView.transcriptMergeDecision(...)`, returning unchanged, append-suffix, or reset-and-append. `mergeNewContent` now uses that helper while preserving the existing reset behavior for shortened or diverged transcripts.

I added direct Swift Testing coverage for append, unchanged, shrink reset, and prefix-divergence reset. Verified with `swift test --package-path native/MuesliNative --scratch-path /private/tmp/muesli-spm-prchunkorder --filter LiveTranscriptMergeDecisionTests` and full `swift test --package-path native/MuesliNative --scratch-path /private/tmp/muesli-spm-prchunkorder` with 1233 tests passing.

ivkiwi:

PR #275, opened 2026-07-05
Add GigaAM v3 Russian ASR backend (CoreML)

ivkiwi:
## What

Adds GigaAM v3 as a local ASR backend — a Russian-first conformer/RNNT model running fully on-device via CoreML. Muesli already covers Russian through multilingual backends (Nemotron, SenseVoice); GigaAM's value is substantially better Russian transcription quality and latency, since it's a Russian-specialized model. In my daily use it clearly outperforms the multilingual options on Russian dictation and meetings.

## Scope (part 1 of 2)

- `GigaAMV3Backend.swift`: model download from Hugging Face (~224 MB CoreML bundles + tokenizer/mel assets), CoreML load (`computeUnits = .all`), file transcription. Long files are transcribed in 20s windows with 2s overlap and suffix-prefix merge (conformer self-attention is O(n²), so unbounded input is both an OOM and quality risk) — chunking is backend-private.
- Registry/routing/UI: backend option, `TranscriptionRuntime` preload/route wiring, model download card following the existing patterns.
- No new package dependencies, no default-model or onboarding changes, no live-meeting pipeline changes.

Part 2 (separate PR, stacked) will add meeting-optimized live chunking (3–20s VAD windows with 2s audio carryover and overlap dedup) — kept out of this PR to keep the live pipeline untouched here.

## Model asset notes

- CoreML bundles: `huggingfinger0/gigaam-v3-coreml` (main); frontend assets: `kruatech/gigaam-v3-mlx` (main)
- Cache: `Application Support/Muesli/Models/gigaam-v3-coreml`, validated by per-file minimum sizes; local seed override via `MUESLI_GIGAAM_V3_MODEL_DIR`

## Tests

File-chunking window math (boundaries/overlap/passthrough/empty), overlap merge on synthetic strings, backend registry/routing coverage. Full suite passes (1223 tests).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
[...]

PR #277, opened 2026-07-05
Tune live meeting chunking for GigaAM (part 2)

ivkiwi:
## What

Part 2 of the GigaAM series (stacked on #275 — its commit appears in this diff until #275 merges; the part-2 delta is `04fd7959`).

GigaAM is a full-utterance conformer: the fixed 3–5s live VAD chunks cut speech mid-word and noticeably degrade live meeting transcription. This PR adds per-backend live chunking configuration: GigaAM meetings use 3–20s VAD windows with 2s of audio carried over between consecutive chunk files, and overlapped text is deduplicated per track. Every other backend keeps the exact current behavior (3–5s, no overlap, no dedup).

## How

- `PCMChunkRecorder`: optional `overlapSampleCount` — each rotated chunk file starts with the last 2s of the previous one; a fresh-bytes guard drops carryover-only chunks.
- `StreamingVadController`: injectable min/max chunk durations (defaults unchanged).
- `MeetingSession`: wires the per-backend config to both mic and system tracks; overlap dedup only when enabled.

## Tests

Byte-identity of carryover across multiple rotations, short-trailing-chunk preservation, per-backend config selection, dedup gating. Full suite passes (1232 tests).

---
<a href="https://app.blacksmith.sh/Muesli-HQ/codesmith/muesli/pr/277"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785867146&installation_id=136549638&pr_number=277&repository=Muesli-HQ%2Fmuesli&return_to=https%3A%2F%2Fgithub.com%2FMuesli-HQ%2Fmuesli%2Fpull%2F277&signature=beca47f7f3f6aca9f43f2646e9f478baa0df0289682003a330e0a04da0c39f7c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added “GigaAM v3 (Russian)” offline speech recognition backend and model UI support.
  * Improved live transcription with backend-tuned VAD chunk durations, overlap-aware chunking, and ordered live emission with per-track overlap deduplication.
* **Bug Fixes**
  * Fixed audio chunk file rotation to correctly carry overlap samples and drop carryover-only chunks.
  * Improved transcript merging/deduplication to reduce repeated words across overlapping segments.
* **Chores**
  * Added/expanded tests covering overlap chunking, live emission, and the new backend.
  * Enhanced download handling with optional progress reporting and safer temp-file cleanup/cancellation.

### DIFF
--- a/native/MuesliNative/Sources/MuesliNativeApp/DownloadUtils.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/DownloadUtils.swift
@@ -20,7 +20,8 @@ func downloadWithRetry(
     from url: URL,
     to destination: URL,
     maxRetries: Int = 3,
-    session: URLSession = .shared
+    session: URLSession = .shared,
+    progressHandler: ((Double) -> Void)? = nil
 ) async throws {
     var lastError: Error?
     for attempt in 0..<maxRetries {
@@ -31,14 +32,17 @@ func downloadWithRetry(
         }
         do {
             try Task.checkCancellation()
-            let (tempURL, response) = try await session.download(from: url)
+            let (tempURL, response) = try await downloadTemporaryFile(
+                from: url,
+                session: session,
+                progressHandler: progressHandler
+            )
             guard let http = response as? HTTPURLResponse, (200...299).contains(http.statusCode) else {
                 let code = (response as? HTTPURLResponse)?.statusCode ?? -1
                 try? FileManager.default.removeItem(at: tempURL)
                 throw DownloadError.httpError(code, url.lastPathComponent)
             }
-            try? FileManager.default.removeItem(at: destination)
-            try FileManager.default.moveItem(at: tempURL, to: destination)
+            try moveDownloadedTemporaryFile(tempURL, to: destination)
             return
         } catch is CancellationError {
             throw CancellationError()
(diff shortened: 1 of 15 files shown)


ivkiwi:

**Greptile inline: PCMChunkRecorder file I/O inside rotation lock**

Fixed in 229a1439. `rotateFile()` now captures carryover samples under the lock, creates and writes the new file outside the lock, then swaps recorder state in a second short lock section. Audio callback appends no longer wait on file creation or carryover I/O.

**Greptile inline: PCMChunkRecorder tailSamples allocation**

Fixed in 229a1439. Tail tracking now mutates `state.tailSamples` in place for the normal case and only allocates a fresh suffix when the incoming buffer alone exceeds the overlap size.

**CodeRabbit inline: DownloadUtils removes destination before move**

Good catch, but this code belongs to the stacked backend commit c1712b42 covered by PR #275, not this live-chunking follow-up. I’m leaving it out of PR #277 so the fix can land with the backend PR.

**CodeRabbit outside diff: refresh chunking config when backend changes**

Fixed in 229a1439 by rejecting `MeetingSession.updateBackend(_:)` once recording is active. Backend changes can still apply before recording starts or on the next meeting, but an active session no longer mixes a new backend with stale chunking, VAD, and dedup settings.

**CodeRabbit nitpick: duplicate download delegate and invalidator**

Agree this is worth cleaning up, but it targets the stacked backend/download work in c1712b42, which belongs to PR #275. I’m not refactoring that path in PR #277 so this follow-up stays limited to live meeting chunking.

**CodeRabbit nitpick: frontend assets sourced from separate Hugging Face repo**

This belongs to the stacked GigaAM backend commit c1712b42 and PR #275. I’m leaving PR #277 focused on live chunking; the asset-source decision should be handled with the backend packaging review.

**CodeRabbit nitpick: Accelerate-backed mel frontend**

Good performance follow-up, but this is in the stacked backend commit c1712b42 for PR #275. I’m not changing the mel frontend in PR #277 so the live-chunking follow-up does not absorb backend implementation work.

**CodeRabbit nitpick: trigram overlap window too large**

Fixed in 229a1439. The trigram and suffix/prefix search window is now capped to 15 words, closer to the configured audio overlap, and there is a regression test for a late repeated trigram that must not drop new words.

**CodeRabbit nitpick: TranscriptOverlapMerger retained full cumulative context**

Fixed in 229a1439. Live dedup now carries only bounded recent overlap context, while `merge(_:)` still returns the full merged transcript for callers that need full text. Added a regression test that keeps live overlap context under 80 words.

**CodeRabbit nitpick: PCMChunkRecorder tailSamples allocation**

Fixed in 229a1439. This is the same hot-path tail buffer issue Greptile flagged; the append path now trims/mutates `tailSamples` without concatenating a full intermediate array.

**Greptile outside diff: previousText accumulator grows without bound**

Fixed in 229a1439. Per-track live overlap state now stores bounded recent context instead of the whole meeting transcript, so long meetings no longer make each chunk emission split all previous words.

ivkiwi:
Follow-up pushed: field testing on our side surfaced two tail-loss defects in the ordered chunk release — a hung chunk could hold all later chunks behind the contiguous gate, and the inactivity-timeout check raced progress append (a chunk completing between an empty poll and the timeout break was dropped). The release path now drains by progress past gaps, failures retire their slot, the timeout check is atomic, and meeting stop drains both collectors with timeout logging. Deterministic regression tests included (50x loop-verified).

ivkiwi:

Issue #267, opened 2026-07-02
Meeting recording should use mp4 instead of wav


javabean:
mp4 files are much more compact than wav.
Saving hours of meeting and audio recording will use a huge amount of disk space in wav form, making the "save audio to disk" unusable in practice.

pHequals7:
https://github.com/Muesli-HQ/muesli/pull/263 - has been fixed - will come out in 0.7.2

PR #278, opened 2026-07-05
Migrate legacy WAV recordings to compact M4A

ivkiwi:
## Problem

The compact M4A writer converts **new** retained recordings, but recordings saved before that change remain full-size WAVs on disk forever (a WAV hour is ~600 MB vs ~60 MB as M4A), and the audio-file import path still persists a converted WAV. Users with meeting history from before the M4A change keep paying the old disk cost. Follows up on the disk-usage concern from #267.

## Fix

- **Startup migration** (background, non-blocking, idempotent): meetings whose saved recording is a WAV on disk are re-encoded to M4A via a temp file + atomic rename; the DB path is updated and the WAV deleted only after a successful encode + DB update, so a quit mid-migration loses nothing. Sub-1KB orphaned `.wav` stubs that no meeting references are cleaned up.
- **Import path** now persists imported recordings as M4A through the same encoder as retained recordings.

## Tests

Migration idempotency, mid-failure safety (encode failure keeps the WAV and DB untouched), orphan stub cleanup, import produces M4A. Full suite passes (1222 tests).

---
<a href="https://app.blacksmith.sh/Muesli-HQ/codesmith/muesli/pr/278"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785879391&installation_id=136549638&pr_number=278&repository=Muesli-HQ%2Fmuesli&return_to=https%3A%2F%2Fgithub.com%2FMuesli-HQ%2Fmuesli%2Fpull%2F278&signature=b44d18b5ff879625f7df51faf6cd5a24c586bf988358b5473b3b56acebd27cc8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a background startup migration to convert legacy WAV meeting recordings to M4A.
  * Imported audio is now persisted as M4A automatically.
* **Bug Fixes**
  * Improved migration reliability with atomic replacement, reference-aware cleanup, and cancellation-safe temporary file handling.
  * Added more robust database opening behavior to reduce “busy” contention during migration.
* **Tests**
  * Expanded async test coverage for migration success, idempotency, shared recordings, and error/cancellation cleanup edge cases.

### DIFF
--- a/REVIEW_RESPONSE.md
+++ b/REVIEW_RESPONSE.md
@@ -1,0 +1,7 @@
+Ready-to-post reply:
+
+Fixed. `migrateLegacyWAVRecordings` now isolates `deleteOrphanedWAVStubs` failures: if the orphan sweep cannot list the recordings directory, it logs that sweep failure, reports `deletedOrphanStubs = 0`, and still returns the migrated count so `startLegacyRecordingMigration` reaches `MuesliNotifications.postDataDidChange()`.
+
+Added regression coverage with an injected `FileManager` whose `contentsOfDirectory` throws for the recordings directory. The test proves a successful WAV-to-M4A migration returns `migrated == 1`, reports `deletedOrphanStubs == 0`, and does not throw.
+
+Validated with `swift test --package-path native/MuesliNative --scratch-path /private/tmp/muesli-spm-prwav --filter MeetingRecordingWriterTests` and `swift test --package-path native/MuesliNative --scratch-path /private/tmp/muesli-spm-prwav`.
--- a/native/MuesliNative/Sources/MuesliCore/DictationStore.swift
+++ b/native/MuesliNative/Sources/MuesliCore/DictationStore.swift
@@ -2756,6 +2756,9 @@ public final class DictationStore {
         if sqlite3_exec(db, "PRAGMA journal_mode=WAL", nil, nil, nil) != SQLITE_OK {
             throw lastError(db)
         }
+        if sqlite3_exec(db, "PRAGMA busy_timeout=5000", nil, nil, nil) != SQLITE_OK {
+            throw lastError(db)
+        }
         return db
     }
 
--- a/native/MuesliNative/Sources/MuesliCore/DictationStore.swift
+++ b/native/MuesliNative/Sources/MuesliCore/DictationStore.swift
@@ -617,14 +617,26 @@ public final class DictationStore {
     }
 
     public func savedRecordingReferenceCount(path: String, excludingMeetingID: Int64? = nil) throws -> Int {
+        try savedRecordingReferenceCount(paths: [path], excludingMeetingID: excludingMeetingID)
+    }
+
+    public func savedRecordingReferenceCount(paths: [String], excludingMeetingID: Int64? = nil) throws -> Int {
         let db = try openDatabase()
         defer { sqlite3_close(db) }
 
-        let standardizedPath = URL(fileURLWithPath: path).standardizedFileURL.path
+        var matchingPaths: [String] = []
+        for path in paths {
+            for candidatePath in [path, URL(fileURLWithPath: path).standardizedFileURL.path] where !matchingPaths.contains(candidatePath) {
+                matchingPaths.append(candidatePath)
+            }
+        }
+        guard !matchingPaths.isEmpty else { return 0 }
+
+        let placeholders = Array(repeating: "?", count: matchingPaths.count).joined(separator: ", ")
         var sql = """
         SELECT COUNT(*)
         FROM meetings
-        WHERE saved_recording_path IN (?, ?)
+        WHERE saved_recording_path IN (\(placeholders))
           AND deleted_at IS NULL
         """
         if excludingMeetingID != nil {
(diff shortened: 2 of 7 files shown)


ivkiwi:

**Finding 1 - CodeRabbit walkthrough and pre-merge summary**

Thanks for the walkthrough. I did not make a code change for this summary-only comment; concrete review findings are handled in the follow-up commit.

**Finding 2 - CodeRabbit inline: migration temp URL ends in `.migrating`**

Fixed. The migration temp filename now keeps `.m4a` as the final extension while retaining a `.migrating` marker, and the writer test asserts the temp destination has the expected M4A extension.

**Finding 3 - CodeRabbit inline: shared WAV path can be deleted too early**

Fixed. Source WAV deletion is now gated by `savedRecordingReferenceCount(path:excludingMeetingID:)`, so the file is removed only after no other active meeting still references it. A focused test covers two meetings sharing one legacy WAV.

**Finding 4 - CodeRabbit nitpick: orphan cleanup should isolate per-file failures**

Fixed. Orphan WAV cleanup now handles each file inside its own `do`/`catch`, logs that file's failure, and continues scanning so one bad entry cannot suppress the migration summary.

**Finding 5 - CodeRabbit nitpick: add SQLite busy timeout**

Added. `openDatabase()` now sets `PRAGMA busy_timeout=5000` after enabling WAL, which gives transient writer contention a retry window without changing migration cancellation or notification behavior.

**Finding 6 - CodeRabbit nitpick: add CancellationError test**

Added. The new cancellation test forces the encoder to create a temp file and throw `CancellationError`, then verifies the error is rethrown, the marker temp file is gone, and the original WAV plus saved recording path remain unchanged.

**Finding 7 - Greptile summary: orphan cleanup can mask successful migrations**

Fixed. This is covered by the same per-file orphan cleanup change: failures are logged for the individual WAV and the migration still returns its summary when earlier migrations succeeded.

**Finding 8 - Greptile inline: orphan `removeItem` failure aborts remaining cleanup**

Fixed. Orphan WAV removal is now inside the same per-file guarded block as file-size and reference-count evaluation, so a single removal failure is logged and the rest of the directory scan continues.

**Finding 9 - Greptile inline: failed source WAV delete has no retry path**

Fixed without adding new state. Orphan cleanup now also deletes an unreferenced WAV when a matching M4A sibling exists, which gives failed post-migration deletes a retry path while avoiding deletion of unrelated large WAV files.

ivkiwi:
Round-2 feedback addressed; follow-up commit pushed.

**Finding 1: Per-file orphan-cleanup failure isolation**

Added coverage for this failure-isolation path. The new test injects a `FileManager` subclass through `migrateLegacyWAVRecordings`, makes `removeItem` throw for one orphan stub, and verifies migration still reports `migrated == 1` plus `deletedOrphanStubs == 2`.

It also verifies the failing stub remains, the other eligible stubs are deleted, and the migrated recording path is still updated.

**Finding 2: Swift Testing throwing expectation**

Updated the cancellation test to use Swift Testing's throwing expectation macro instead of manual success/failure bookkeeping.

Confirmed the chosen async throwing expectation compiles and passes with the package's pinned Swift Testing version.

ivkiwi:
Round-3 finding addressed; follow-up commit pushed.

Fixed in 8778aec109eb636bc0b530f4d6deaec17dea6ac2. The legacy WAV migration now reads file attributes inside the per-file do/catch, so metadata failures for one source are logged and skipped without aborting the remaining migration batch.

Added a regression test with a FileManager subclass that throws attributesOfItem for one WAV path and verifies the next WAV still migrates, with summary counts remaining migrated == 1 and deletedOrphanStubs == 0.

Validated with the targeted MeetingRecordingWriter test and the full MuesliNative SwiftPM suite using scratch /private/tmp/muesli-spm-prwav.

ivkiwi:
Round-4 finding addressed; follow-up commit pushed.

Fixed in 901bf0c2edddb86b37d439c26984e08beaf2fabf.

The migration now counts saved recording references across both the raw legacy path from the candidate row and the standardized source path before deleting the source WAV. The shared store helper expands each supplied path to its standardized form and deduplicates matches before querying, so raw and standardized legacy rows are counted together.

Added a regression where two sibling meetings store the same shared WAV through a raw non-standardized path. The test verifies both meetings migrate to the M4A path before the original WAV is removed.

Validation passed: `swift test --package-path native/MuesliNative --scratch-path /private/tmp/muesli-spm-prwav --filter MeetingRecordingWriterTests` and `swift test --package-path native/MuesliNative --scratch-path /private/tmp/muesli-spm-prwav`.

ivkiwi:
Round-5 finding addressed; follow-up commit pushed.

Fixed. `migrateLegacyWAVRecordings` now isolates `deleteOrphanedWAVStubs` failures: if the orphan sweep cannot list the recordings directory, it logs that sweep failure, reports `deletedOrphanStubs = 0`, and still returns the migrated count so `startLegacyRecordingMigration` reaches `MuesliNotifications.postDataDidChange()`.

Added regression coverage with an injected `FileManager` whose `contentsOfDirectory` throws for the recordings directory. The test proves a successful WAV-to-M4A migration returns `migrated == 1`, reports `deletedOrphanStubs == 0`, and does not throw.

Validated with `swift test --package-path native/MuesliNative --scratch-path /private/tmp/muesli-spm-prwav --filter MeetingRecordingWriterTests` and `swift test --package-path native/MuesliNative --scratch-path /private/tmp/muesli-spm-prwav`.

ivkiwi:

PR #281, opened 2026-07-06
Cancel scheduled starting-now prompts once the user acts

ivkiwi:
## Problem

The upcoming-meeting notification schedules a per-event "Meeting starting now" timer, but none of the action closures — Record, Join & Record, Join Only, Dismiss — cancel it. After the user acts on the notification, a redundant "Meeting starting now" prompt still fires at event start time (visible whenever the recording ends before the scheduled start, or after a plain dismiss).

## Fix

The notification key is passed into the handler so every action closure cancels and removes its `meetingStartingNowTimers` entry.

## Tests

Regression coverage for all four action paths. Full suite passes (1220 tests).

---
<a href="https://app.blacksmith.sh/Muesli-HQ/codesmith/muesli/pr/281"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785918326&installation_id=136549638&pr_number=281&repository=Muesli-HQ%2Fmuesli&return_to=https%3A%2F%2Fgithub.com%2FMuesli-HQ%2Fmuesli%2Fpull%2F281&signature=8eef64d8e896542b232630691908ecd5de4151557367c25688bd5f4102f15df3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated upcoming meeting notification handling so the correct “starting now” reminder timer is canceled when you take an action (start recording, join with recording, join only, or dismiss).
  * Helps prevent duplicate or outdated reminder prompts after responding to a meeting notification.
* **Tests**
  * Added a unit test to verify user actions cancel the corresponding “starting now” timer before proceeding.

### DIFF
--- a/native/MuesliNative/Sources/MuesliNativeApp/MuesliController.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/MuesliController.swift
@@ -2313,7 +2313,7 @@ final class MuesliController: NSObject {
             )
 
             // Show "starts in X min" notification now
-            handleUpcomingMeeting(upcomingEvent)
+            handleUpcomingMeeting(upcomingEvent, notificationKey: key)
 
             // Schedule a second "Meeting starting now" notification at event start time for pre-start prompts.
             let delay = event.startDate.timeIntervalSinceNow
@@ -7479,7 +7479,7 @@ final class MuesliController: NSObject {
         }
     }
 
-    private func handleUpcomingMeeting(_ event: UpcomingMeetingEvent) {
+    private func handleUpcomingMeeting(_ event: UpcomingMeetingEvent, notificationKey: String? = nil) {
         // Look up end date and meeting URL from unified calendar events
         let calendarEvent = appState.upcomingCalendarEvents
             .first(where: { $0.id == event.id })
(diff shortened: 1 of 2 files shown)


ivkiwi:
Thanks for the reviews — addressed below.

**Finding 1: Default `nil` on `handleUpcomingMeeting`**

Good catch. Fixed in eaaac034 by making `handleUpcomingMeeting` require `notificationKey: String` and making `cancelMeetingStartingNowTimer` take a non-optional key. Future call sites now get a compiler-visible requirement instead of silently bypassing cancellation.

**Finding 2: Source-text parsing test did not verify runtime behavior**

Fixed in eaaac034. I replaced the source-text parsing test with a runtime check around `ScheduledPromptActionPolicy.performUserAction`, which verifies the captured notification key is passed to cancellation before the user action runs. The scheduled prompt action closures now route through that helper.

**Finding 3: `muesliControllerSource()` path reconstruction was brittle**

Fixed in eaaac034. The test no longer reconstructs `MuesliController.swift` from `#filePath` or parses source sections. It exercises the cancellation helper directly, so directory layout changes will not break this coverage.

ivkiwi:

PR #284, opened 2026-07-06
Chunk long files for SenseVoice transcription

ivkiwi:
## Problem

`SenseVoiceBackend` passes the entire audio file to FluidAudio's `SenseVoiceManager` in a single call. The manager builds one input tensor over all samples and clamps internally, so a long imported recording silently transcribes only its opening seconds — the rest of the audio is dropped with no error — while also allocating a very large transient tensor.

## Fix

Long files are transcribed app-side in ~15s windows with a short overlap, sequentially, and the window texts are merged with suffix-prefix overlap deduplication (backend-private helper). Short files keep the existing single-call path unchanged.

## Tests

Window math (boundaries, overlap, passthrough, empty input) and merge dedup on synthetic strings. Full suite passes (1220 tests).

---
<a href="https://app.blacksmith.sh/Muesli-HQ/codesmith/muesli/pr/284"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785918879&installation_id=136549638&pr_number=284&repository=Muesli-HQ%2Fmuesli&return_to=https%3A%2F%2Fgithub.com%2FMuesli-HQ%2Fmuesli%2Fpull%2F284&signature=ce543471e86300134bfac6dbf89be4caef6130c5279d6e55c5d4f9d449d8f151"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Improved transcription for longer audio by automatically splitting it into overlapping chunks based on detected duration.
  * Enhanced transcript merging to reduce duplicated text and improve stitching across spaced and unspaced languages, including special handling for CJK and alphanumeric boundaries.
  * Added more accurate audio duration detection and reliable window-based audio sample reading to support chunked transcription.
* **Tests**
  * Added coverage for chunking decisions, overlap-based transcript merging behavior, and audio window reading correctness.

### DIFF
--- a/native/MuesliNative/Sources/MuesliNativeApp/SenseVoiceBackend.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/SenseVoiceBackend.swift
@@ -1,5 +1,82 @@
+import AVFoundation
 import FluidAudio
 import Foundation
+
+enum SenseVoiceFileChunking {
+    static let sampleRate = SenseVoiceConfig.sampleRate
+    static let passthroughThresholdSeconds: TimeInterval = 15
+    static let windowSeconds: TimeInterval = 15
+    static let overlapSeconds: TimeInterval = 2
+
+    static func shouldChunk(duration: TimeInterval) -> Bool {
+        duration > passthroughThresholdSeconds
+    }
+
+    static func shouldChunk(sampleCount: Int, sampleRate: Int = sampleRate) -> Bool {
+        guard sampleCount > 0, sampleRate > 0 else { return false }
+        return shouldChunk(duration: Double(sampleCount) / Double(sampleRate))
+    }
+
+    static func windows(sampleCount: Int, sampleRate: Int = sampleRate) -> [Range<Int>] {
+        guard sampleCount > 0 else { return [] }
+        guard shouldChunk(sampleCount: sampleCount, sampleRate: sampleRate) else {
+            return [0..<sampleCount]
+        }
+
+        let windowSamples = max(1, Int((windowSeconds * Double(sampleRate)).rounded()))
+        let overlapSamples = min(Int((overlapSeconds * Double(sampleRate)).rounded()), windowSamples - 1)
+        let stepSamples = windowSamples - overlapSamples
+        var result: [Range<Int>] = []
+        var start = 0
+
+        while start < sampleCount {
+            let end = min(start + windowSamples, sampleCount)
+            result.append(start..<end)
+            if end == sampleCount { break }
+            start += stepSamples
+        }
+
+        return result
+    }
+
+    static func mergeTranscripts(_ transcripts: [String]) -> String {
+        SenseVoiceTranscriptMerger.merge(transcripts)
+    }
+}
+
+private enum SenseVoiceTranscriptMerger {
+    private static let maxOverlapWords = 40
+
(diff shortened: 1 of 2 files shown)


ivkiwi:

**Greptile PR summary**

Addressed both risks in `f88a46cb8cdf36090bb3bebb418eaee125b5d7ce`. The long-file path no longer keeps the full resampled audio array alive; it reads and converts each SenseVoice window from the source file before transcribing that window. I also added CJK character-overlap merging so unspaced Chinese/Japanese/Korean transcript boundaries deduplicate instead of falling through the whitespace-token path.

**Inline comment 3527475268**

Fixed in `f88a46cb8cdf36090bb3bebb418eaee125b5d7ce`. The chunked SenseVoice path now creates windows from the target sample count, then reads and converts only the current window from `AVAudioFile` before sending it to FluidAudio. That removes the full-recording `[Float]` lifetime from the long-import path.

**Inline comment 3527475329**

Fixed in `f88a46cb8cdf36090bb3bebb418eaee125b5d7ce`. The merger now detects unspaced CJK scripts and uses character-level suffix-prefix overlap for those transcripts. The existing word-based merge path remains in place for space-delimited transcripts.

**Review 4634176670**

No reply needed for the empty review body. The actionable inline comments from this review were handled separately.

ivkiwi:
Round-2 feedback addressed; follow-up commit pushed.

**Finding 1: CodeRabbit inline `3527578853`**

Fixed in `0cd3043f`. Character-mode merging now preserves a boundary space when a zero-overlap chunk join would otherwise connect two ASCII alphanumeric tokens. The guard is limited to ASCII letters/numbers, so pure CJK chunk joins keep the previous separator-free behavior.

I added a mixed CJK/ASCII regression test covering the glued-token case.

**Finding 2: CodeRabbit review `4634300770`, trailing window lower bound**

Fixed in `0cd3043f`. The bounded-window test now checks the trailing window lower bound as well as the upper bound, so a near-empty trailing read would fail.

**Finding 3: CodeRabbit review `4634300770`, duplicated overlap helpers**

Leaving this unchanged for this pass. The two overlap helpers are small and intentionally type-specific, while a generic helper would add abstraction without addressing a behavioral issue. I kept the follow-up scoped to the functional merge and test-coverage findings.

**Finding 4: Greptile edited summary, sample-rate wording**

No code change here. The chunked path does not derive window ranges from the duration helper; it opens `SenseVoiceAudioWindowReader` and builds ranges from the reader sample count, which already uses the processing format sample rate. The duration helper only gates the short-file passthrough path and labels the log line for the WAV input path.

ivkiwi:

PR #304, opened 2026-07-10
Notion sync: one-way export of meetings and dictations to Notion

### LINKED ISSUE #305: Sync meetings and dictations to Notion
### Motivation

Muesli captures meeting notes, transcripts, and dictations locally, but many people organize their knowledge in Notion. Today there's no way to get Muesli's output into a Notion workspace without manual copy-paste. Tools like Granola have set the expectation that meeting notes flow into Notion automatically.

### Proposed behavior

A one-way sync that pushes meetings (AI notes + transcript) and dictations into a database in the user's **own** Notion workspace. Muesli stays the source of truth; Notion is a mirror.

- **Opt-in**, off by default, configured in Settings → Sync → Notion.
- **Local-first / no backend.** Uses a Notion **internal integration token** the user pastes, so there's no Muesli server in the path and no OAuth backend to run. The user shares one page with the integration; Muesli creates a styled `"<Name>'s Muesli Notes"` database under it (banner, icon, welcome). A second Mac reuses the same database.
- **Automatic, event-driven.** A meeting/dictation syncs shortly after it's created or edited. Editing re-syncs the same page (no duplicates). Deleting in Muesli archives the Notion page; deleting in Notion never touches Muesli.
- **Readable pages.** Each record is a row (Type / Date / Meeting type / Created by); the body is the notes as Notion blocks with the transcript in a collapsible toggle.
- **Privacy.** Audio is never sent — only notes, transcripts, and metadata. The token is stored in a hardened local file (0600, excluded from backups), never in `config.json` or iCloud. One-way: Muesli never reads the workspace.

### Non-goals / follow-ups

- Two-way sync (Notion → Muesli).
- Capturing meeting **attendees** as a property (meetings don't store calendar attendees yet) — a follow-up.
- OAuth "Connect to Notion" one-click flow (would require a backend to hold the client secret; deliberately out of scope for a local-first app).

### Implementation

Implemented in #304.

ahamino:
Closes #305

## Summary

Adds a **Notion integration**: a one-way sync that pushes your meetings (AI notes + transcript) and dictations into a Notion database in **your own** workspace. Muesli stays the source of truth; Notion is a mirror. It's local-first — there's **no Muesli backend** in the path, just your Mac talking to the Notion API with a token you own.

Opt-in, off by default, configured in **Settings → Sync → Notion**.

## How it works

**Setup.** You paste a Notion **internal integration token** and share a page with the integration (••• → Connections). On enable, Muesli creates a single styled database — `"<Your name>'s Muesli Notes"` (named from your onboarding name) — under that shared page, with a Muesli banner, icon, and a short welcome. A second machine reuses the same database instead of duplicating it.

**Syncing is event-driven.** A few seconds (debounced) after you create or edit a meeting/dictation, it's pushed. Each meeting becomes a row with `Type` / `Date` / `Meeting type` / `Created by` properties; the page body is the notes rendered as Notion blocks (headings, nested bullets) with the raw transcript tucked in a collapsible toggle. Editing re-syncs the **same** page (no duplicates); deleting in Muesli **archives** the Notion page. Deleting in Notion never affects Muesli (the sync only ever writes).

**Data model.** Per-record export state lives in a `publish_state_json` column keyed by target (`{"notion":{"id":…,"s":…}}`), so adding a future export target needs no schema change. The remote page id is persisted the instant the page is created (idempotent on retry) and replicated via CloudKit so multiple Macs converge on one Notion page.

## Design notes

- **Internal token, not OAuth.** Notion's REST OAuth requires a `client_secret` (no secretless PKCE for this API surface), which would force a backend — at odds with Muesli's no-server design. The internal token keeps it fully local; the cost is a one-time manual page-share, which the UI walks you through. (Confirmed live: internal tokens can't create at the workspace root, so the DB lives under the shared page.)
- **Notion 2026-03-11 data-sources + Markdown API.** Page bodies are sent as Notion-flavored Markdown, so there's no hand-rolled block converter to maintain — Notion renders headings, nested lists, and `<details>` toggles server-side.
- **Hardened token storage.** The token lives in a dedicated `notion-auth.json` (POSIX `0600`, excluded from backups), mirroring how `ChatGPTAuthManager` stores OAuth tokens — never in the general `config.json`, never synced to iCloud.
- **Pluggable `ExportTarget` layer.** The engine (enumerate dirty records → render → push/update/unpublish → mark) is target-agnostic; Notion is the first implementer. Adding e.g. Obsidian later is a new conformer, not a rewrite.

## Changes

Four commits, by layer:

1. **Storage** (`DictationStore`, `StorageModels`) — `publish_state_json` column + migration, export/unpublish queries, transactional writes with `busy_timeout`, CloudKit replication of publish-state.
2. **Client + engine** (`NotionClient`, `ExportTarget`/`ExportCoordinator`, `NotionTarget`, `NotionCredentialStore`, `MuesliICloudSyncEngine`) — the REST client, the pluggable export layer (idempotent id persistence, drain loop, surfaced failures, delete propagation), Notion rendering, hardened token file.
3. **Controller + UI** (`MuesliController`, `SettingsView`, `AppState`, `Models`) — setup/reuse under a shared page, share validation, debounced+drained push that re-runs edits landing mid-push, token lifecycle, the Settings section.
4. **Tests** — coordinator idempotency/drain/delete-propagation, Markdown body/payload shapes, store publish-state round-trip + queries, credential store + config decode, DB-title derivation.

No changes outside the Notion feature; the config migration is additive (new snake_case keys default for old configs).

## Testing

- **Live-validated** end-to-end against a real Notion workspace (create styled DB under a shared page → markdown records with nested bullets + transcript toggle → in-place re-sync → archive-on-delete).
- `swift build` clean; **672 tests pass** across the store, sync, CloudKit, and Notion suites.
- Went through a full multi-angle code review; the correctness findings (busy_timeout, non-atomic re-sync, cross-device dedup, lifecycle guards, MainActor IO) are fixed and folded in.

## Privacy

Audio is never sent. Only notes, transcripts, and the metadata above are written. One-way (Muesli never reads the workspace). The token stays on the Mac in a permission-locked, backup-excluded file.

## Known limitations / follow-ups

- **Cross-device dedup** (CloudKit replication of the page id) is logically sound but can only be fully exercised with two Macs on one iCloud account — worth a real two-device check before release.
- **Attendees** on each record are not yet captured (meetings don't store calendar attendees today) — parked as a follow-up pipeline.
- Re-sync of a very long page clears + re-inserts blocks (bounded by Notion's ~3 req/s); a native diff/replace would be cheaper if Notion adds one.

## Summary by CodeRabbit

* **New Features**
  * Added one-way Notion sync for meetings and dictations, including an in-app **Sync now** control.
  * Automatically creates/reuses a **“Muesli Notes”** Notion database and persists per-record remote IDs for idempotent retries.
  * Introduced robust one-way unpublishing/deletion propagation with export-state tracking per destination.
* **Bug Fixes**
  * Prevented clobbering existing Notion page links during cloud sync and improved concurrent write stability.
* **Documentation**
  * Documented the one-way export target architecture for adding new destinations.
* **Tests**
  * Expanded coverage for Notion token/config handling, export/unpublish behavior, and migration/state persistence.


ahamino:
Thanks — all six findings were valid; addressed in `ff5b82ee`.

**CodeRabbit**
- **Migration SELECT drops `publish_state_json`** ✅ Confirmed and fixed. `textDictationRecordsForSyncMigration`/`…Meeting…` now select `publish_state_json` as the last column, matching the index the shared mapper reads (12/13) — the page id no longer gets read out-of-range as nil during the CloudKit seed. Added a regression test (`textRecordsForSyncMigration` preserves the page id).
- **Retrying non-idempotent writes** ✅ Fixed. `send()` now retries **429/529 for all methods** (request wasn't processed) but restricts **5xx/transport retries to idempotent methods** (GET/HEAD/DELETE); POST/PATCH throw immediately, so a lost response can't duplicate a page/database/markdown write.

**Greptile**
- **P1 — unescaped pagination cursor** ✅ Fixed. `pageChildBlockIDs` now percent-encodes the opaque `next_cursor` before interpolating it into the query. (The other paginators pass the cursor in a JSON body, so they were unaffected.)
- **P1 — disable race** ✅ Fixed. Disabling sync now cancels+nils `notionPushDebounceTask` and the in-flight `notionPushTask`, and a `notionPushGeneration` guard prevents a superseded run's completion from clobbering a re-enabled push. The coordinator also checks `Task.isCancelled`, so a cancelled push halts promptly instead of finishing the batch.
- **P2 — asset URLs on `main`** ✅ Fixed. Icon/banner now pin to an immutable commit SHA so a future rename can't break icons on already-created databases.
- **P2 — `RelativeDateTimeFormatter` per render** ✅ Fixed. Hoisted to a `static let`.

Build clean; 174 tests pass.

pHequals7:
Hey @ahamino, thanks for working on this - just want to confirm if this is using an official Notion API key for the integration? I am not too aware of the current state of notion's API ecosystem - also we have a post-meeting hook that is configurable by the end user to run automations once a meeting is over - how well does this feature re-use that primitive?

ahamino:
Thanks @pHequals7 — good questions.

**Official API key?** Yes. It uses an official **Notion internal-integration token** against the REST API (version `2026-03-11`, the data-sources + Markdown API). The user generates the token from their own Notion integration and pastes it in; it stays fully on-device in a hardened `notion-auth.json` (`0o600`, backup-excluded) and never touches a Muesli server. I deliberately avoided OAuth here: Notion's OAuth REST flow requires a `client_secret` (there's no secretless PKCE for the REST API — only their MCP surface has PKCE), which would force us to stand up and host a token-exchange backend. That's incompatible with the local-first, no-server design, and it would also mean users' meeting content routes through our infra. The internal-token model keeps everything local and user-owned.

**Reuse of the post-meeting hook?** It doesn't reuse it, and that's intentional rather than an oversight. The two solve different problems:

- `MeetingHookRunner` is a fire-and-forget escape hatch — on `meeting.completed` it shells out to a user executable with a minimal stdin payload (`{schemaVersion, event, kind, id, completedAt}`). It's perfect for "run *my* arbitrary automation," but it hands the script only an id; the script owns everything after that.
- Notion sync is a first-class, stateful integration. It needs things the id-only hook can't carry: a durable local→remote page-id mapping, idempotent create-vs-update (so a re-run never duplicates a page), in-place re-sync when notes change, delete propagation (Muesli→Notion archive), retry/backoff, and CloudKit cross-device dedup so two Macs don't double-publish. That's why it's built as a pluggable `ExportTarget` + `ExportCoordinator` driven off local-change state, not a one-shot completion callback.

So the relationship is: the hook is the general-purpose "bring your own automation" primitive; Notion is a built-in target that maintains sync state the hook was never meant to hold. They're complementary — a power user could still wire their own Notion script via the hook, but they'd be reimplementing all of the above by hand.

Happy to reconcile them if you'd prefer a tighter story — e.g. dispatch the Notion push off the same `meeting.completed` trigger, or expose `ExportTarget` more broadly so other built-in destinations ride the same coordinator. Just let me know which direction you'd want.

pHequals7:
understood - would this notion sync integration be unique to notion alone or does this pave away for muesli to integrate with other third party services so that in the future we can have an integration platform of sorts to connect other services as well? if you can structure your PR in such a way would help future contributors..

ahamino:
Great question! 😄 Nope — it's not Notion-only. Notion is just the **first** target riding on a general one-way export layer.

The whole "hard part" is service-agnostic: there's an `ExportTarget` protocol and an `ExportCoordinator` that handles dirty-record tracking, idempotent create-vs-update, delete propagation, retries/backoff, auth reconnect prompts, and cross-device dedup. Export state lives in a generic `publish_state_json` column keyed by target, so a new service needs **zero schema changes**.

So adding, say, Obsidian or a webhook is really just: implement `ExportTarget` (a client + a render step) and register it — you inherit everything else for free. 🎉

One honest note: it's a one-way **export/mirror** framework (Muesli → service), not a two-way or hosted-OAuth "integration hub" — keeping it local-first was the priority.

And totally agree on helping future contributors — so I went ahead and added a short guide in this PR: **[`docs/export-targets.md`](https://github.com/Muesli-HQ/muesli/blob/feature/notion-sync/docs/export-targets.md)** (linked from `CONTRIBUTING.md`). It covers the protocol contract, what the coordinator gives you for free, the single wiring point, and a ~40-line stub target as a worked example. Hopefully that makes the next service a fill-in-the-blanks job! 🙌

ahamino:
🤖 Babysitter pass — pushed one fix to `feature/notion-sync` (`beda389`).

**Fixed & pushed**

- **Duplicate-page bug on re-sync** (`NotionSyncEngine.swift`) — this was the one outstanding blocking finding in Greptile's summary (flagged as still-unaddressed at 4/5 confidence). In `NotionTarget.export()`, the child-block delete loop lived inside the same `do/catch` that treats `isNotFound`/`isArchived` as *"the page is gone → recreate it."* So if a user moved or deleted a single child block in Notion, its 404 during the delete loop fired the outer catch **after** fresh Markdown had already been inserted into the still-existing page — falling through to `createPage` and producing a duplicate. Each per-block `deleteBlock` now has its own catch that tolerates a missing/archived block (matching the existing `unpublish()` pattern), so only a genuine page-level not-found from the property/markdown writes triggers recreation.

**Deliberately left for a human**

- **No regression test for this path.** `NotionTarget.export()` calls the `static NotionClient` methods directly (no injectable seam), and the coordinator tests exercise stub `ExportTarget`s rather than `NotionTarget`. Covering the child-block-404 case would mean introducing a client seam — an architectural change beyond the scope of this fix. Worth doing as a follow-up if you want this pinned down by a test.
- **CodeRabbit "Docstring coverage 11.86%" pre-merge warning** — a threshold/style nit, not a correctness issue; left untouched.
- **@pHequals7's questions** (official Notion token, post-meeting-hook reuse, general integration-platform direction) look answered and are product/direction calls, not code changes — leaving those to you and the maintainers.

Note: I couldn't compile or run the suite here (Linux runner, no Swift toolchain — this is a macOS/AppKit target). The change mirrors an already-proven pattern in the same file, but please run `swift test --package-path native/MuesliNative` on macOS before merge.

I did **not** merge or approve — that's the maintainers' call.

PR #320, opened 2026-07-14
Fix: microphone capture dies silently when the default input device changes mid-recording

spacevibes365:
## Problem

During a meeting recording, if the default audio input device changes mid-recording — the most common case being AirPods connecting — the microphone side of the recording dies **silently**. System audio keeps flowing, the meeting completes with `status: completed`, and the user only discovers later that their own speech is missing from both the saved WAV and the transcript. No warning is shown.

### Reproduction (observed on v0.7.1, macOS 26.5.2, M4 Pro)

1. Start a meeting recording with built-in speakers + built-in mic (play any video as the "remote" side).
2. ~3.5 minutes in, connect AirPods (they become the default input & output).
3. Speak several times while wearing AirPods.
4. Take AirPods off, stop the recording, save.

**Result:** the saved recording and transcript contain only the system-audio side after the switch. Re-transcribing the saved mixed WAV independently confirmed zero mic-side speech after the AirPods connect. `MeetingMicHealthTracker` did not surface a warning in this scenario.

### Root cause

`StreamingMicRecorder` drives the mic through a long-lived `AVAudioEngine`, but nothing in the codebase observes `AVAudioEngineConfigurationChange` (`grep` finds zero hits). When the engine's I/O configuration changes, AVAudioEngine stops delivering input buffers to the installed tap, and the recorder never notices: `isRunning` stays true, the file handle stays open, and silence (nothing at all, in fact) is appended for the rest of the meeting.

`RouteAwareMeetingMicRecorder` selects a recorder at start time and reacts to `preferredInputDeviceID` changes, but a *system default* input change mid-recording never reaches it.

## Fix

- Observe `.AVAudioEngineConfigurationChange` for the recorder's engine while a recording is active (installed after a successful `engine.start()`, removed in `stop()`/`cancel()`).
- On change, serialized on a dedicated queue and under the existing `graphLock`: remove the tap, stop the engine, reset the prepared-graph state, re-run `prepareLocked()` (which re-applies `preferredInputDeviceID`), rebuild the tap + `AVAudioConverter` against the **new** hardware format, and restart the engine.
- Capture keeps appending to the **same output file** — the tap's target format is a constant 16 kHz mono, so the WAV stays valid and the recording continues with only the gap that the OS itself introduced.
- If the restart fails, the error is reported through the existing `onRecordingFailed` path (same as tap-level failures), and `stop()` still finalizes the partial file.

The extraction of `startEngineWithTapLocked(recordingID:)` from `start()` is behavior-preserving; `start()`'s failure cleanup is unchanged.

Since `StreamingMicRecorder` also backs the dictation recorders, hands-free dictation sessions survive input-device changes too.

## Testing

- `swift build` (release) and the full `swift test` suite pass — 1397/1399; the 2 failing assertions are the pre-existing locale-dependent `ContributionMilestone` share-message test (grouping separator on a `ru_RU` machine: `31 000` vs `31,000`), unrelated to this change.
- Manual verification of the repro scenario on a patched dev build is in progress; I'll post the before/after logs (`[streaming-mic] engine configuration changed; restarting input capture`) here.

## Possible follow-ups (out of scope)

- Surface a user-visible notice when a mic restart happens or fails (`MeetingMicHealthTracker` currently misses the "engine stopped delivering callbacks entirely" case if system audio is also quiet).
- `CoreAudioSystemRecorder.restartTapForDefaultOutputDeviceChange()` has a similar silent-failure path on the system-audio side (`isRecording = false` + stderr log only).

---
<a href="https://app.blacksmith.sh/Muesli-HQ/codesmith/muesli/pr/320"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786654180&installation_id=136549638&pr_number=320&repository=Muesli-HQ%2Fmuesli&return_to=https%3A%2F%2Fgithub.com%2FMuesli-HQ%2Fmuesli%2Fpull%2F320&signature=e184ddcec94307fb3016e08c9cb4acdc4cfa2b9166bb8fd6c41be9230198d911"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated “Meetings microphone” selector in Settings > Meetings, including an “Automatic” option.
  * Changing the Meetings microphone during an active meeting now applies immediately to meeting capture.

* **Bug Fixes**
  * Improved microphone recording stability when input devices or audio routing change mid-recording, including seamless recovery.
  * Recording now resumes into the same audio file after an internal audio engine restart, preserving continuity.
  * Strengthened start/stop/pause handling to prevent incomplete output files, and improved microphone health warnings when capture fails.


spacevibes365:
Addressed both review-bot findings in f1e73b3: the failed-restart path in `handleEngineConfigurationChange()` now removes the tap, stops the engine, and unregisters the configuration-change observer (so a recording can't silently resume after `onRecordingFailed` was delivered), and `deinit` removes the observer as a safety net. Package builds clean and the test suite passes (the two `ContributionMilestone` failures on my machine are pre-existing locale-dependent number formatting, unrelated to this change).

spacevibes365:
Follow-up hardening in 68041d9: a deeper review pass found that removing the observer on a failed restart did not cover callbacks already queued on the serial queue — a second queued notification could still resume capture after `onRecordingFailed`, and a stale callback from a stopped recording could restart the next one. The observer now captures the recording ID and the handler only proceeds when that ID is still active and no failure has been reported (same pattern as `captureGeneration` in `AudioQueueInputRecorder`). The observer is also installed before `engine.start()` so a device change landing in that window can't reproduce the original silent failure. Build clean, full suite passes apart from the pre-existing locale-dependent `ContributionMilestone` case.

pHequals7:
Resolved through #333, merged into main as signed merge commit f373f309c06b89b103d76e98cad521fd4fa2b998. #333 preserves @spacevibes365's original three commits and adds the scoped handoff, recovery, diagnostics, CI coverage, and successful AirPods transition validation. GitHub automatically marked this PR merged when its commits reached main through #333. Thank you for identifying and implementing the original configuration-change recovery.

PR #335, opened 2026-07-22
Fix crash in dictation correction monitor on non-finite AX geometry

### LINKED ISSUE #323: Crash after dictation in cmux when Dictionary suggestions scans AX geometry
## Summary

With **Dictionary suggestions** enabled, completing a dictation in **cmux** consistently crashes Muesli immediately after transcription becomes ready.

The Left Option hotkey itself works: recording starts, audio is captured, and recording stops normally. The crash occurs when `DictationCorrectionMonitor` walks cmux's accessibility tree after dictation.

## Environment

- Muesli: **0.8.0** (build 0.8.0)
- macOS: **26.5.2 (25F84)**
- Hardware: MacBook Pro, Apple M4 Max (`Mac16,5`, arm64)
- cmux: **0.64.18** (build 98, `com.cmuxterm.app`)
- Dictation hotkey: **Left Option** (key code 58, 250 ms hold threshold)
- Dictionary suggestions: **enabled**
- Accessibility permission: granted

## Steps to reproduce

1. Enable **Dictionary suggestions** in Muesli.
2. Focus a text-input surface inside cmux.
3. Hold Left Option long enough to begin dictation.
4. Speak a short phrase.
5. Release Left Option and wait for transcription to finish.

## Expected behavior

Muesli inserts the dictation and begins monitoring for a user correction without crashing.

## Actual behavior

Muesli crashes within the same second that dictation reaches `ready_for_transcription`, then relaunches.

This reproduced **8 times across July 14-15**. Each matching crash corresponded to a completed dictation whose `app_context` was `cmux|com.cmuxterm.app`. Dictations into VS Code, ChatGPT/Codex, WhatsApp, Screen Sharing, and Chrome during the same period did not produce this stack.

The matching reports terminate with:

```
Exception: EXC_BREAKPOINT (SIGTRAP)
Queue: com.apple.root.utility-qos.cooperative

specialized static DictationCorrectionMonitor.elementVisitKey(_:)
specialized static DictationCorrectionMonitor.collectTextSnapshots(...)
specialized static DictationCorrectionMonitor.collectTextSnapshots(...)
specialized static DictationCorrectionMonitor.collectTextSnapshots(fromProcessID:add:)
static DictationCorrectionMonitor.detectEditedSnapshots(originalText:targetApp:)
closure in DictationCorrectionMonitor.start(...)
```

The final latency events before the crash are:

```
threshold_met:hold-start
first_buffer
first_speech
stop_finished
ready_for_transcription
<crash>
```

## Likely cause

Current main builds the accessibility visit key by directly converting AX geometry to `Int`:

https://github.com/Muesli-HQ/muesli/blob/9e090f2862cd264b90dd4f38ebdacfbb9c368822/native/MuesliNative/Sources/MuesliNativeApp/DictationCorrectionMonitor.swift#L1259-L1270

```swift
if let position, let size {
    return "\(pid)|\(role)|\(Int(position.x))|\(Int(position.y))|\(Int(size.width))|\(Int(size.height))"
}
```

Accessibility geometry is untrusted. A cmux AX element appears to return a non-finite or out-of-`Int`-range coordinate/dimension. Swift traps when converting values such as `NaN`, infinity, or an out-of-range `CGFloat` to `Int`. That matches both `EXC_BREAKPOINT/SIGTRAP` and the top crash frame in `elementVisitKey`.

## Possible fixes

1. Validate every coordinate and dimension with `isFinite` and an explicit `Int` range check before conversion. If any value is invalid, use the existing `CFHash(element)` fallback.
2. Alternatively, use `pid + role + CFHash(element)` for the visit key unconditionally and remove geometry conversion from this path.
3. Extract the geometry-to-key logic into a pure helper and add regression cases for:
   - `CGFloat.nan`
   - positive and negative infinity
   - `CGFloat.greatestFiniteMagnitude`
   - normal negative/off-screen coordinates
4. Treat malformed AX attributes as a skipped/fallback node and continue scanning; an accessibility element should never be able to terminate the process.

For example:

```swift
private static func safeInt(_ value: CGFloat) -> Int? {
    guard value.isFinite,
          value >= CGFloat(Int.min),
          value <= CGFloat(Int.max)
    else { return nil }
    return Int(value)
}

if let position,
   let size,
   let x = safeInt(position.x),
   let y = safeInt(position.y),
   let width = safeInt(size.width),
   let height = safeInt(size.height) {
    return "\(pid)|\(role)|\(x)|\(y)|\(width)|\(height)"
}

return "\(pid)|\(role)|\(CFHash(element))"
```

## Suggested regression coverage

- A unit test for safe key generation with non-finite and out-of-range geometry.
- A manual/integration smoke test: dictation into cmux with Dictionary suggestions enabled must complete without a crash.
- Confirm correction monitoring still detects edits after falling back from invalid geometry.

## Workaround

Disabling **Dictionary suggestions** avoids `DictationCorrectionMonitor` and prevents this crash while leaving dictation itself usable.

Gui11aum3:
I've had a few crashes in the last version (v0.8.0) when running dictation, the app crashes.
This is my attempt at fixing this based on the app bug report.

## Summary

Fixes a crash (`EXC_BREAKPOINT` / `SIGTRAP`) in `DictationCorrectionMonitor.elementVisitKey` triggered after dictation, while walking the target app's accessibility tree.

The Accessibility framework can return non-finite **or out-of-range** geometry — `NaN`, `±Infinity`, or finite-but-enormous values — for offscreen, placeholder, or degenerate AX nodes. Swift's `Int(_: Double)` initializer traps on all of these.

Every `CGFloat` → `Int` conversion is now guarded with `Int(exactly:)` applied to `.rounded(.towardZero)`, so any element whose geometry can't be represented exactly falls through to the existing `CFHash`-based visit key instead of crashing.

Fixes #323

## Root cause

Crash report registers (x8, x19, x24) all held `0x7ff0000000000000` — IEEE-754 `+Infinity`. The AX tree walker reached an element with infinite geometry and the interpolation `Int(position.x)` trapped.

## Implementation

| Commit | Author | Change |
|---|---|---|
| `49e8ccf8` | @Gui11aum3 | Initial fix — `.isFinite` guards on position/size |
| `c4f8f72a` | @Gui11aum3 | **Operative fix** — `Int(exactly:)` + `.rounded(.towardZero)` |
| `45a06727` | @pHequals7 | Regression coverage — no behavior change |

`c4f8f72a` is the fix that matters. `.isFinite` alone was insufficient: `.greatestFiniteMagnitude` *is* finite but still traps in `Int()`. `Int(exactly:)` closes `NaN`, `±Infinity` and out-of-range in a single construct, while `.rounded(.towardZero)` preserves the original truncation semantics — visit keys stay byte-identical for normal geometry, so tree-walk dedup is unchanged.

`45a06727` extracts the logic into `nonisolated static func elementVisitKey(processID:role:position:size:fallbackHash:)`, taking plain values rather than an opaque `AXUIElement` (which can't be constructed with arbitrary geometry in a unit test), and adds the regression suite. Behavior-preserving.

## Testing

- `swift test --package-path native/MuesliNative` — **1432 tests across 140 suites, all passing**
- `git diff --check` — clean
- New `Dictionary correction monitor` suite (4 tests):
  - truncation toward zero for normal geometry
  - negative / offscreen coordinates
  - fallback when position or size is missing
  - fallback for `NaN`, `+∞`, `−∞`, `.greatestFiniteMagnitude` — independently across x, y, width and height

The `.greatestFiniteMagnitude` case fails on `49e8ccf8` and passes on `c4f8f72a`, so it pins the checked-conversion approach against a future revert to the `.isFinite` form.

### Runtime verification

The patched build was installed as the default `MuesliDev` lane and exercised across repeated cmux dictations — no crash, process survived throughout. Dictionary correction prompts were separately confirmed working in AX-capable apps.

**The original crash was not reproduced.** #323 was reported against cmux `0.64.18` (build 98); verification here ran on cmux `0.64.20` (build 100), with identical macOS (26.5.2 / 25F84) and Muesli (0.8.0). A pre-fix baseline build also did not crash, so there is no before/after demonstration. The guard is validated by regression tests against the trapping condition rather than by reproducing the original fault. Note the reporter observed 8 crashes across two days of ordinary use, so the per-dictation rate is unknown and a single clean baseline run is weak evidence either way.

### Known limitation — not addressed here

Dictionary correction prompts do not appear in cmux, before or after this change. cmux's terminal AX surface exposes only currently-selected text, never full field contents, so the correction monitor has nothing to diff. This is orthogonal to the crash — #323's "expected behavior" wording touches it, but a crash guard was never going to deliver it. Tracking separately as follow-up.

---
<a href="https://app.blacksmith.sh/Muesli-HQ/codesmith/muesli/pr/335"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787320897&installation_model_id=422865&pr_number=335&repository=Muesli-HQ%2Fmuesli&return_to=https%3A%2F%2Fgithub.com%2FMuesli-HQ%2Fmuesli%2Fpull%2F335&signature=9ba03d52236de8bb5d57bc5ff83502373e963a7a6f32fb41594d45eadf0fac14"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary

* **Bug Fixes**
  * Improved dictation correction monitoring by validating accessibility elements’ position and size data before using coordinate-based tracking.
  * When geometry values can’t be safely represented, the system now uses a consistent fallback identifier to keep monitoring reliable.

* **Tests**
  * Added coverage to verify visit-key formatting across valid/negative/offscreen geometry and invalid floating-point values.

* **Chores**
  * Updated CI test sharding so the new dictation-transcription test suite runs in the appropriate shard.

### DIFF
--- a/native/MuesliNative/Sources/MuesliNativeApp/DictationCorrectionMonitor.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/DictationCorrectionMonitor.swift
@@ -1264,14 +1264,30 @@ final class DictationCorrectionMonitor {
         let position = cgPointAttribute(kAXPositionAttribute, from: element)
         let size = cgSizeAttribute(kAXSizeAttribute, from: element)
 
+        return elementVisitKey(
+            processID: pid,
+            role: role,
+            position: position,
+            size: size,
+            fallbackHash: CFHash(element)
+        )
+    }
+
+    nonisolated static func elementVisitKey(
+        processID: pid_t,
+        role: String,
+        position: CGPoint?,
+        size: CGSize?,
+        fallbackHash: CFHashCode
+    ) -> String {
         if let position, let size,
            let px = Int(exactly: position.x.rounded(.towardZero)),
            let py = Int(exactly: position.y.rounded(.towardZero)),
            let sw = Int(exactly: size.width.rounded(.towardZero)),
            let sh = Int(exactly: size.height.rounded(.towardZero)) {
-            return "\(pid)|\(role)|\(px)|\(py)|\(sw)|\(sh)"
+            return "\(processID)|\(role)|\(px)|\(py)|\(sw)|\(sh)"
         }
-        return "\(pid)|\(role)|\(CFHash(element))"
+        return "\(processID)|\(role)|\(fallbackHash)"
     }
 
     nonisolated private static func axElementAttribute(_ attribute: String, from element: AXUIElement) -> AXUIElement? {
--- a/native/MuesliNative/Sources/MuesliNativeApp/DictationCorrectionMonitor.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/DictationCorrectionMonitor.swift
@@ -1265,9 +1265,11 @@ final class DictationCorrectionMonitor {
         let size = cgSizeAttribute(kAXSizeAttribute, from: element)
 
         if let position, let size,
-           position.x.isFinite, position.y.isFinite,
-           size.width.isFinite, size.height.isFinite {
-            return "\(pid)|\(role)|\(Int(position.x))|\(Int(position.y))|\(Int(size.width))|\(Int(size.height))"
+           let px = Int(exactly: position.x.rounded(.towardZero)),
+           let py = Int(exactly: position.y.rounded(.towardZero)),
+           let sw = Int(exactly: size.width.rounded(.towardZero)),
+           let sh = Int(exactly: size.height.rounded(.towardZero)) {
+            return "\(pid)|\(role)|\(px)|\(py)|\(sw)|\(sh)"
         }
         return "\(pid)|\(role)|\(CFHash(element))"
     }
--- a/native/MuesliNative/Sources/MuesliNativeApp/DictationCorrectionMonitor.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/DictationCorrectionMonitor.swift
@@ -1264,7 +1264,9 @@ final class DictationCorrectionMonitor {
         let position = cgPointAttribute(kAXPositionAttribute, from: element)
         let size = cgSizeAttribute(kAXSizeAttribute, from: element)
 
-        if let position, let size {
+        if let position, let size,
+           position.x.isFinite, position.y.isFinite,
+           size.width.isFinite, size.height.isFinite {
             return "\(pid)|\(role)|\(Int(position.x))|\(Int(position.y))|\(Int(size.width))|\(Int(size.height))"
         }
         return "\(pid)|\(role)|\(CFHash(element))"
--- a/native/MuesliNative/Tests/MuesliTests/DictationCorrectionMonitorTests.swift
+++ b/native/MuesliNative/Tests/MuesliTests/DictationCorrectionMonitorTests.swift
@@ -1,0 +1,87 @@
+import Foundation
+import Testing
+@testable import MuesliNativeApp
+
+@Suite("Dictionary correction monitor")
+struct DictationCorrectionMonitorTests {
+    private let processID: pid_t = 42
+    private let role = "AXTextArea"
+    private let fallbackHash: CFHashCode = 9_001
+
+    @Test("visit key preserves truncation toward zero for finite geometry")
+    func visitKeyPreservesTruncationTowardZero() {
+        let key = DictationCorrectionMonitor.elementVisitKey(
+            processID: processID,
+            role: role,
+            position: CGPoint(x: 12.9, y: -3.7),
+            size: CGSize(width: 640.8, height: -20.2),
+            fallbackHash: fallbackHash
+        )
+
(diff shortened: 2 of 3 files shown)


pHequals7:
Thanks @Gui11aum3  - this is when the dictionary suggestions are turned on I presume? Been working on a fix - will take a look at this as well.

PR #371, opened 2026-08-03
Fix media never resuming after dictation

yxssxn:
## The bug

With `pause_media_during_dictation` enabled, media pauses when dictation starts and then never resumes. Once that has happened the feature degrades further: the user resumes playback by hand and subsequent dictations behave inconsistently, and in some cases stop pausing altogether until the app is restarted.

Reproduced on macOS 26.5.2 with Plexamp. The cause is not player-specific.

## Root cause: MediaRemote reports success for commands it does not apply

`SystemMediaPlaybackClient.sendMediaCommand` dispatched explicit `pause` / `play` commands through `MRNowPlayingController.localRouteController` and treated the call returning without error as proof the command took effect.

Measured directly against a genuinely playing Plexamp session on macOS 26.5.2, sending each command and then re-reading the now-playing playback rate:

| Command | Reported | Observed effect |
|---|---|---|
| `sendCommand(1)` — pause | success | paused |
| `sendCommand(0)` — play | success | none |
| `sendCommand(2)` — togglePlayPause | success | none |

So the pause path happened to work and the resume path silently did nothing. Because the JXA route reported success, the media-key fallback underneath it was never reached.

I also probed `MRMediaRemoteSendCommand` directly from a locally-signed helper binary in the same session: it returned `true` for `pause`, `play` and `togglePlayPause`, and none of the three changed playback state, while the `localRouteController` `pause` above worked in the same session. That is one player on one macOS version, so I would not state it more strongly than that — but it does mean a `true` return from either API is not evidence a command landed.

There is a second, independent reason explicit commands cannot be relied on here: Electron players — Plexamp among them — bind media keys through `globalShortcut` and register no MediaRemote command handlers at all, so no MediaRemote command reaches them regardless of how it is dispatched.

## The fix

Use the `NX_KEYTYPE_PLAY` HID event — the one the physical F8 key produces — as the sole command transport. Native and Electron players both honour it.

That event is a toggle with no acknowledgement, so it cannot be sent blindly; sending one at the wrong moment is exactly what #225 was about. Every transition is therefore verified:

1. Query the now-playing state.
2. If it already matches the goal, send nothing.
3. Otherwise send one toggle.
4. Poll until the system confirms the new state.
5. If it never confirms, the event was lost — send one more, then give up.

State is never guessed at. An `unknown` reading is treated as "do not touch", so an unreadable state signal can never start silent media.

Reading now-playing state through MediaRemote does still work, which is why the query is shelled out through `osascript`; the existing in-process `MRMediaRemoteGetNowPlayingApplicationIsPlaying` is retained as a fallback that returns `unknown` rather than a wrong answer.

## Also fixed: a latched pause that disabled the feature permanently

`beginDictationMediaPause` returned early whenever the controller was already in its `paused` state. If a dictation ever ended without a restore — cancelled, errored, or a hotkey release that never arrived — the controller stayed latched in `paused` and silently stopped pausing for every subsequent dictation until the app restarted.

Begin now treats a pre-existing `paused` state as stale, re-arms, and re-checks live playback state. This is what lets the feature recover on its own.

## Other changes

- Media restore no longer waits behind ducking restore. It was nested inside the ducking completion handler, which delayed resume even when ducking was disabled for the session and had nothing to undo.
- Restore no longer sends a command when the user resumed playback by hand during dictation — verify-before-toggle skips it rather than toggling playback back off.

## Second commit, separable

`Pause media during dictation on headphone routes too` is a behaviour change rather than part of the resume fix, so it is a separate commit and can be dropped without affecting the rest. `beginDictationMediaPause` returned early on any non-speaker route, so dictating on headphones produced no pause at all. That route check belongs to ducking, which exists to stop speaker bleed reaching the microphone; pausing serves a different purpose. Happy to drop it if you would rather keep the route gate.

## Resume obligations survive session boundaries

A resume starts with a state query, so a dictation beginning inside that window — about one query, ~70ms measured — cancels it before any toggle is sent. The next begin then reads `notPlaying`, correctly declines to pause, and nothing afterwards owns the resume, leaving media paused indefinitely. Same outcome if the resume toggle is sent but lost and the retry is then cancelled.

`owesResume` is therefore tracked separately from session state: set when we pause, cleared only when a play transition is *confirmed*. A restore honours an outstanding obligation even when the current session never paused anything.

Covered by `dictationDuringInFlightResumeDoesNotStrandMedia`, which holds the resume query open, starts a second session inside the window, and asserts media ends up playing. Verified to fail without the fix.

Credit to CodeRabbit for catching this on review; I had wrongly assumed the cancelled case still left media playing.

## Known limitation: media-key routing

macOS routes the play/pause key to one app, and that target is not always the app MediaRemote reports as now-playing. Measured here: with Plexamp running but its queue empty, Plexamp still held the key target and swallowed the event, so a toggle aimed at audio playing in Chrome did nothing — MediaRemote reported `Google Chrome` as now-playing throughout. Quitting Plexamp made the same toggle work immediately.

The controller degrades correctly rather than misbehaving: the toggle is retried, the state never confirms, `settled=false` is logged, and no spurious resume is issued afterwards because the restore path re-reads state and finds it already playing. Media simply keeps playing through the dictation.

Separately, QuickTime Player registers as now-playing but honours neither the HID key nor any MediaRemote command, so it cannot be paused by any remote mechanism. Also handled by the same degradation path.

Both are macOS/player behaviours rather than anything this code can route around, but they are worth knowing when reading `settled=false` in the logs.

## Testing

`MediaPlaybackControllerTests` rewritten against a fake client that models a real player: a toggle flips playback state, and toggles can be dropped to simulate lost HID events. 14 tests covering the retry path, the in-flight-resume path, the self-heal path, the already-paused no-op (#225 regression guard), the unknown-state no-op, and repeated pause/resume cycles.

- Full suite: 1484 tests across 145 suites pass.
- Verified in-app on macOS 26.5.2 across two players: five consecutive real dictation cycles against Plexamp (Electron), and five more against a Chrome media session (Blink), each confirming `pause settled=true` then `play settled=true` with playback sampled mid-hold.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved media playback restoration during dictation, allowing playback to resume without waiting for audio ducking restoration.
  * Prevented unnecessary play/pause commands when playback is already in the intended state.
  * Added verification and retry handling for missed playback commands.
  * Improved recovery from unavailable or delayed playback-state information.
* **Tests**
  * Expanded coverage for playback restoration timing, repeated dictation cycles, delayed responses, dropped commands, and headphone playback.

### DIFF
--- a/native/MuesliNative/Sources/MuesliNativeApp/DictationAudioSessionManager.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/DictationAudioSessionManager.swift
@@ -537,8 +537,12 @@ final class DictationAudioSessionManager: @unchecked Sendable {
     }
 
     private func restoreSessionAudioState(completion: (() -> Void)? = nil) {
-        duckingController.restoreDictationDucking { [self] in
-            self.mediaPlaybackController.restoreDictationMediaPause()
+        // Resuming the user's media is independent of ducking cleanup and must
+        // not queue behind it. Nesting it inside the ducking completion delayed
+        // the resume by however long ducking restore took — including when
+        // ducking was disabled for the session and had nothing to undo.
+        mediaPlaybackController.restoreDictationMediaPause()
+        duckingController.restoreDictationDucking {
             completion?()
         }
         routingController.refreshRouteAfterDictationSession()
--- a/native/MuesliNative/Sources/MuesliNativeApp/MediaPlaybackController.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/MediaPlaybackController.swift
@@ -7,12 +7,30 @@ protocol MediaPlaybackManaging: AnyObject {
     func restoreDictationMediaPause()
 }
 
+/// What we are trying to achieve, not how it is delivered. The only transport
+/// that works across every player is the hardware play/pause key, which is a
+/// *toggle*; the intent is what decides whether that toggle should be sent at
+/// all, and what state we then verify against.
+enum MediaPlaybackCommand: String, Equatable, CustomStringConvertible {
+    case pause
+    case play
+
+    var description: String { rawValue }
+
+    var targetState: MediaPlaybackState {
+        switch self {
+        case .pause: return .notPlaying
+        case .play: return .playing
+        }
+    }
+}
+
 /// Actual playback state of the system now-playing application, as opposed to
 /// `AudioOutputActivityStatus`, which only reflects whether an app's audio
 /// output pipeline is running. Browsers and most video players keep their
 /// audio engine/IO alive while a video is *paused*, so `IsRunningOutput` (and
 /// therefore `AudioOutputActivityStatus`) reports "active" for paused media
-/// and cannot be used to decide whether to send a play/pause toggle.
+/// and cannot be used to decide whether to send a media command.
 enum MediaPlaybackState: Equatable, CustomStringConvertible {
     case playing
     case notPlaying
@@ -31,51 +49,81 @@ protocol MediaPlaybackClient {
     /// Whether the current now-playing application is actually producing audio.
     /// `.unknown` is returned when the signal cannot be obtained.
     func nowPlayingPlaybackState(completion: @escaping (MediaPlaybackState) -> Void)
-    func sendMediaPlayPauseToggle()
+
+    /// Sends one hardware-equivalent play/pause toggle. `intent` is diagnostic
+    /// only — the delivered event is identical in both directions, which is why
+    /// callers must confirm the resulting state rather than assume it.
+    func sendPlayPauseToggle(intent: MediaPlaybackCommand)
 }
 
+/// Pauses whatever is playing for the duration of a dictation and puts it back
+/// afterwards.
+///
+/// Every transition is *verified*: query the now-playing state, send a toggle
+/// only if the state disagrees with the goal, then poll until the system
+/// confirms the new state, retrying the toggle once if it was dropped. Media
+/// commands on macOS report success far more readily than they take effect, so
+/// an unverified send is indistinguishable from a no-op. See
+/// `SystemMediaPlaybackClient` for why the toggle is the only usable transport.
 final class MediaPlaybackController: MediaPlaybackManaging {
+    struct VerificationPolicy {
+        /// Total toggles per transition, including the first.
+        var maxToggleAttempts: Int
+        /// State polls per toggle before that toggle is considered lost.
+        var pollsPerAttempt: Int
+        var pollInterval: DispatchTimeInterval
+
+        static let `default` = VerificationPolicy(
+            maxToggleAttempts: 2,
+            pollsPerAttempt: 12,
+            pollInterval: .milliseconds(80)
+        )
+    }
+
     private enum PauseState: Equatable {
         case idle
+        /// Query in flight; we have not yet touched playback.
         case checkingBegin(Int)
-        case paused
-        case checkingRestore(Int)
+        /// We sent a pause for this session and owe a resume.
+        case paused(Int)
     }
 
     private let client: MediaPlaybackClient
     private let queue: DispatchQueue
+    private let policy: VerificationPolicy
     private var pauseState: PauseState = .idle
     private var generation = 0
+    private var inFlight = 0
 
     init(
         client: MediaPlaybackClient = SystemMediaPlaybackClient(),
-        queue: DispatchQueue = DispatchQueue(label: "com.muesli.media-playback")
+        queue: DispatchQueue = DispatchQueue(label: "com.muesli.media-playback"),
+        policy: VerificationPolicy = .default
     ) {
         self.client = client
         self.queue = queue
+        self.policy = policy
     }
 
     func beginDictationMediaPause(enabled: Bool, routeKind: AudioOutputRouteKind) {
         queue.async { [self] in
             guard enabled else { return }
             guard routeKind == .speakerLike else { return }
+
             switch pauseState {
-            case .idle:
-                generation += 1
-                let token = generation
-                pauseState = .checkingBegin(token)
-                client.nowPlayingPlaybackState { [weak self] playbackState in
-                    self?.queue.async {
-                        self?.handleBeginPlaybackState(playbackState, token: token)
-                    }
-                }
-            case .checkingRestore:
-                // A new dictation began before the previous restore query
-                // completed. Keep the media paused and let this new session own
-                // the eventual restore instead of briefly resuming playback.
-                pauseState = .paused
-            case .checkingBegin, .paused:
+            case .checkingBegin:
+                // A query for this session is already in flight.
                 return
+            case .paused:
+                // A previous session never got its restore — a cancelled or
+                // errored dictation, or a release we never saw. Without this,
+                // the controller would sit in `.paused` forever and silently
+                // stop pausing for every future dictation.
+                MediaPlaybackLogging.log("begin found stale paused state; re-arming")
+                pauseState = .idle
+                startBeginPlaybackCheck()
+            case .idle:
+                startBeginPlaybackCheck()
             }
         }
     }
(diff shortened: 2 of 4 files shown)


yxssxn:
Thanks — the in-flight resume finding is real and now fixed in 72bef13b.

One correction to the analysis: the window is not the full ~1.9s verification period. `ensure(.play)` sends its toggle immediately after the first state query returns, so once the toggle is out, a cancelled verification only loses the retry, not the resume. The actual hole is narrower — roughly one state query (~70ms measured) between `restoreDictationMediaPause` and the toggle being sent.

The consequence you identified is right regardless, and worse than I had it in the PR description. I had assumed the cancelled case still left media playing; it does not. If the toggle was never sent, the next begin reads `notPlaying`, correctly declines to pause, and nothing afterwards owns the resume — so media stays paused indefinitely. That is the exact failure this PR targets, so I have stopped treating it as an acceptable limitation.

Rather than defer the new begin or extend the token guard, the fix tracks the obligation separately from session state: `owesResume` is set when we pause and cleared only when a play transition is *confirmed*. A restore honours an outstanding obligation even when the current session never paused anything. That also covers the related case where the resume toggle is sent but lost, and the retry is then cancelled.

Added `dictationDuringInFlightResumeDoesNotStrandMedia`, which holds the resume query open with `completesImmediately: false`, starts a second session inside the window, and asserts media ends up playing. Verified it fails without the source change (`client.playbackState → not-playing`) and passes with it. Full suite 1484 tests.

yxssxn:
Both review findings are real and fixed in 8ac90173.

**Greptile P1 — pause verification left active on release.** Correct, and my earlier "obligation" fix did not cover it. `restoreDictationMediaPause` reused the pause session's generation token, so a release arriving mid-verification left the pause verifier authorised to keep retrying. With a dropped first toggle that retry lands after the resume and playback ends stopped; the oscillation you describe is the same mechanism. Restore now bumps the generation before starting the resume, retiring the pause verifier. `owesResume` still carries the obligation, so retiring the pause cannot strand playback — the two fixes compose rather than conflict.

Added `releaseRetiresPauseVerification`, which stalls the verification queries, drops the first toggle, and releases mid-flight. Verified it fails without the source change: a second `.pause` toggle fires and playback ends `not-playing`.

**CodeRabbit — query log volume and now-playing application name.** Also fair. Verification polls the state up to twelve times per transition and every call logged a line naming the now-playing application. The success path is now silent; only a backend fallback logs, since that indicates the preferred path is unavailable and is worth knowing. The application name is no longer logged, parsed, or returned by the JXA script at all — the controller only ever needed the playback rate to make a decision, so there was no reason to read it.

Full suite 1485 tests. Also re-verified in-app across five consecutive real dictation cycles after these changes.

yxssxn:
Third finding, also correct, fixed in d08f7f4e.

Confirmed reachable in `DictationAudioSessionManager`: `stop`, `cancel`, `endExternalSession` and the failure path all call `restoreSessionAudioState`, and `cancel` sets its state unconditionally with no guard against an already-stopped session — so two restores for one dictation is an ordinary sequence, not a corner case.

`resumeIfOwed` started a second verifier on the current generation, so both stayed live, both read paused media, and the later toggle stopped playback again. Measured without the guard: **four** play toggles fire as the racing verifiers each retry, and playback ends `not-playing`. Worse than the two-toggle case described.

The active resume is now tracked by token and a duplicate restore declines to start a second. The one already running finishes the job; if it is cancelled by a new dictation, `owesResume` outlives it and the next restore retries. The comparison is by token rather than a bare flag, so a superseded resume finishing late cannot clear the marker belonging to the resume that replaced it.

`duplicateRestoreDoesNotRaceFirstResume` stalls the first resume, fires a second restore, and asserts exactly `[.pause, .play]`. Verified to fail without the guard.

That is three concurrency findings in this controller from review — the in-flight-resume cancellation, the pause/resume verifier overlap, and now duplicate restores. They share one root: verification is asynchronous, so every path that can start or cancel a transition needs to say which transition it owns. The invariants now are: at most one live verifier per direction, generation retires superseded work, and `owesResume` is the only thing that survives cancellation. Full suite 1486 tests.

I will confirm in-app across real dictation cycles and report back rather than assert it here.

yxssxn:
In-app verification, as promised: **6/6 consecutive real dictation cycles** on a build of d08f7f4e, each pausing on press and resuming on release.

The useful number is toggles per cycle: **exactly 2** every time, one pause and one play. That is the direct counter-evidence for the racing-verifier failures — the duplicate-restore bug produced four play toggles in the unit test, and the verifier-overlap bug produced a second pause. Neither appears in practice now.

Log volume is 4 lines per dictation (`toggle intent=pause`, `pause settled=true`, `toggle intent=play`, `play settled=true`), down from roughly 20, with no application name.

Verified against a Chrome media session rather than Plexamp this time, which also confirms the fix is not Electron-specific — Plexamp exercised the same path in earlier rounds.

Two macOS behaviours worth recording for anyone reading `settled=false` in the logs, both already noted in the PR description and neither addressable in this code:

- QuickTime Player registers as now-playing but honours neither the HID key nor any MediaRemote command.
- The play/pause key target is not always the app MediaRemote reports as now-playing. With Plexamp running but idle, it held the key target and swallowed events aimed at audio playing in Chrome; quitting Plexamp made the same toggle work immediately.

In both cases the controller retries, fails to confirm, logs `settled=false`, and issues no spurious resume — media simply keeps playing through the dictation.

yxssxn:
Fixed in 4e9e9944 — and this one was a regression I introduced in the previous commit, so thank you for staying on it.

The duplicate-restore guard blocked on *any* active resume token, including one belonging to a superseded generation. Your sequence is exactly right: a new dictation retires the in-flight resume, ends before that stale callback runs, `resumeIfOwed` sees the stale token and declines, then the callback clears the token without resuming — leaving `owesResume` set with nothing left to act on it. Media stays paused indefinitely, which is precisely the failure the obligation exists to prevent.

The guard now blocks only on a resume for the *current* generation, the one that can still make progress. A superseded resume is already abandoned and must not hold the obligation hostage.

`supersededResumeDoesNotBlockNextRestore` reproduces the ordering by resolving only the newer session's query and leaving the retired resume's callback outstanding — the fake client grew a `completeLast` for this, since `completeNext` could not express "resolve the newer one while the older stays pending". Verified to fail without the change, and `duplicateRestoreDoesNotRaceFirstResume` verified still to pass with it, so the two constraints coexist rather than trade off.

For the re-flagged **Duplicate restores race resume toggles** comment at line 165: that is the finding addressed in the previous commit (d08f7f4e). If it is still reachable in the current code I would rather know than assume it is a stale re-render, so please flag it again if so.

Full suite 1487 tests. Will confirm in-app and report back.

yxssxn:
In-app verification of 4e9e9944: **6/6 consecutive real dictation cycles**, exactly 2 toggles each, one pause and one play.

Toggle count is the measure that matters here — the racing-verifier defects each produced extra toggles (four play toggles for the duplicate-restore case, a second pause for the verifier overlap), so a steady 2 is direct evidence none of them fire in practice.

Full suite 1487 tests. Five regression tests now cover this controller, each individually verified to fail without its fix:

- `dictationDuringInFlightResumeDoesNotStrandMedia`
- `releaseRetiresPauseVerification`
- `duplicateRestoreDoesNotRaceFirstResume`
- `supersededResumeDoesNotBlockNextRestore`
- `droppedToggleIsRetried` / `droppedResumeToggleIsRetried`

The invariants they pin down, stated plainly since four of the six defects here were concurrency rather than transport:

1. At most one live verifier per direction.
2. `generation` retires superseded work; a retired verifier must neither act nor block.
3. `owesResume` is the only state that survives cancellation, and is discharged only on confirmation.

Happy to squash before merge if you would prefer a cleaner history — the commits are currently split so the headphone-route behaviour change stays separable.

yxssxn:
Rebased onto current `main` (`aaa4b49c`). The branch was 334 commits behind; it is now up to date and conflict-free. The net diff is unchanged from the previously reviewed state, only replayed onto a newer base.

The workflow runs from the 3 August push expired unapproved after the 30-day window, so CI never executed on this branch. The rebase queued fresh runs, which are again waiting on maintainer approval:

- CI: https://github.com/Muesli-HQ/muesli/actions/runs/33640993009
- Claude Code Review: https://github.com/Muesli-HQ/muesli/actions/runs/33640993010

Could a maintainer approve the runs so the checks can complete? DCO, both Socket Security checks and Greptile have already passed on this head.

Prior review rounds on this PR were addressed and verified in-app: 6/6 consecutive dictation cycles with exactly 2 toggles per cycle, one pause and one play.

Issue #303, opened 2026-07-09
Meeting detection incorrectly ends Teams recording after brief periods of silence


tonylex:
Hi! First of all, thank you for building Muesli. It's been a fantastic tool and I'm really enjoying using it.

I believe I've found a reproducible issue with automatic meeting detection in Microsoft Teams.

## Environment

- macOS (Apple Silicon M5)
- Latest version of Muesli
- Microsoft Teams desktop application (not the browser version)

## Steps to reproduce

1. Join a Microsoft Teams meeting.
2. Wait for the **"Meeting detected"** notification.
3. Click **Start Recording** from the notification.
4. Recording starts normally.
5. During the meeting, if everyone is silent for a short period (for example while reading code, reviewing a document, or waiting for someone), Muesli automatically stops recording.
6. It immediately starts generating the meeting summary.
7. The Teams meeting is still active and nobody has left the call.
8. As soon as someone starts speaking again, Muesli displays **"Meeting detected"** again as if it were a brand new meeting.

## Expected behavior

The recording should continue until the Teams meeting actually ends or until the user manually stops the recording. Short periods of silence during a meeting should not cause the recording to end.

## Actual behavior

Muesli appears to interpret a brief period of silence as the end of the meeting. It stops recording, starts summarizing the transcript, and then immediately detects the same ongoing Teams meeting again when someone resumes speaking.

## Additional observations

- This issue is easy to reproduce and happens consistently.
- It only occurs when recording is started from the **"Meeting detected"** notification.
- If I manually click **"Start Meeting"** inside Muesli before or during the meeting, the recording works correctly and continues until I manually stop it.

pHequals7:
Interesting bug; the meeting detection is a function of mic and camera activity, will need to repro this on my mac and get back to you - thanks for flagging..

federicobechini:
Same problem here, with one extra piece of root-cause evidence from the macOS unified log that might help the repro.

In my case the trigger is not actual silence. It is a **transient reconfiguration of the input USB device** (a Logitech Brio 300, which is camera + microphone on a single USB device). At every point where Muesli ended the session, the log shows:

```
AppleUSBAudioEngine:...:Brio 300:...: Abandoning I/O cycle because reconfig pending
HALC_ProxyIOContext::IOWorkLoop: ending the transport, stopping the io thread
setPlayState Stopped Input {com.muesli.system-audio-tap-...}
```

[...]

PR #374, opened 2026-08-03
fix: stop meeting recording only when the stop control is clicked

jamesagarside:
## Summary

The floating indicator draws a **6pt** stop dot 8pt from the trailing edge, but the click handler treated **every point past `x=30`** as a stop:

```swift
if x < 30 { ...pause... } else { ...stop... }   // stop == the entire rest of the pill
```

On a recording pill up to 190pt wide, that made the clickable stop region roughly **25x larger** than the affordance it represents. Any click on the pill body — repositioning the indicator, dismissing the hover transcript, or an accidental click on a window that sits above everything else — silently finalized a live meeting and sent it straight to summarization. No confirmation, and no log line, which also made it undiagnosable after the fact.

I found this while chasing intermittently truncated meetings on my own install (0.8.1): several sessions ending at 31–160s with healthy transcripts right up to the cut. Two controlled runs isolated it — recording from a terminal-launched build so stderr was captured:

- **Run 1**, normal mouse use: died at 52s. No `[meeting] auto-stopping…` marker, no `[streaming-mic] engine configuration changed` marker, no capture failure. A completely clean teardown, i.e. something invoked `stopMeetingRecording()` directly.
- **Run 2**, identical but hands off the mouse entirely: **survived**, still recording past 118s.

The only controlled difference was the pointer.

### What changed

`FloatingIndicatorControlLayout` now owns the control geometry, and both the drawing code (`addStopLayer`, the recording-state `iconLabel` frame) and click hit-testing read from it, so the visible control and its clickable region cannot drift apart again. Each control gets a **24pt minimum horizontal touch target** spanning the full pill height; clicks on the body are a no-op. Dragging to reposition is unchanged.

### Behaviour change worth calling out

This applies to dictation as well as meetings. Body-clicking the pill during a toggle dictation previously stopped it and now does not — the same drawn affordance governs both, so I've kept them consistent rather than special-casing. Happy to scope it to meetings only if you'd rather minimise the change.

### Related

Related to but distinct from #303. That issue is the detection-driven auto-stop path (`MeetingAutoStopTracker` / signal-loss timeout); this touches none of it. Both can truncate a live meeting, which made them easy to confuse — the tell is that the auto-stop path logs and this one did not.

### Note on CI coverage

`FloatingIndicatorPointerInteractionTests` was on the `ci_unsharded_test_suites.txt` legacy allowlist, so it has **never run in CI**. I've assigned it to the `core` shard and removed the stale allowlist entry; `scripts/test_ci_test_shards.sh` passes.

## Validation

- `swift build --package-path native/MuesliNative` — clean.
- `swift test --filter FloatingIndicatorPointerInteractionTests` — 7 tests pass. Includes a case asserting that every x from 30→150 (the whole of the old stop region) now resolves to `.body`, plus full-pill-height reachability and narrow-pill overlap resolution.
- `bash scripts/run_ci_test_shard.sh core` — **244 tests / 16 suites, all passing**.
- `bash scripts/test_ci_test_shards.sh` — shard assignments verified.
- The existing `singleClickStillRuns` test pinned the old behaviour (`handleClick(atX: 50)` expecting a stop); it is replaced by explicit stop-control and body-click cases.

## Contribution certification

      under Muesli's
      and licenses or terms.
      resulting changes.

### Third-party materials

None.

### AI assistance

Claude Code was used to investigate the root cause, implement the change, and write the tests. All changes were reviewed and tested locally, including the two controlled reproduction runs described above.

---
<a href="https://app.blacksmith.sh/Muesli-HQ/codesmith/muesli/pr/374"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788383568&installation_model_id=422865&pr_number=374&repository=Muesli-HQ%2Fmuesli&return_to=https%3A%2F%2Fgithub.com%2FMuesli-HQ%2Fmuesli%2Fpull%2F374&signature=83422892c3b2ac202aeaddefb385504d21adb6f8db2b6ebe820b20a3981b289c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recording pill interactions by distinguishing clicks on the leading control, trailing stop control, and central body.
  * Clicking the pill body no longer unintentionally stops recording.
  * Controls remain accessible across the pill’s full height and narrow layouts.
  * Programmatic stop actions continue to work as expected.

* **Tests**
  * Expanded coverage for recording indicator hit areas, control targeting, narrow-pill behavior, and click handling.


jamesagarside:
Thanks — good catch, and it was a real gap rather than a style nit.

Confirmed the arithmetic: for a 36pt pill the touch targets are `[0, 24]` (leading) and `[13, 37]` (trailing), overlapping at `[13, 24]`. My two probe points were the control centres at x=12 and x=25, both of which sit *outside* that overlap, so each resolved through a single-target branch and the `(true, true)` tie-break was never covered.

Fixed in a2fc4c80 by probing either side of the boundary between the two control centres (x=18.5), which is inside the overlap.

I verified the new assertions genuinely reach the branch by mutation testing — inverting the tie-break comparison to `? .leadingControl : .trailingControl` fails **only** the two new assertions, while the two original centre-point assertions still pass. That reproduces your analysis exactly.

PR #375, opened 2026-08-03
fix: stop tearing down meeting recordings when the observed identity changes

jamesagarside:
## Summary

`MeetingAutoStopPolicy.matches()` compares a live candidate against the armed source by **candidate ID**, **suppression ID**, and **normalized URL**. It never consults the source's bundle ID.

`MeetingMediaSessionTracker` mints a fresh session ID once its 30s quiet window lapses. A dedicated meeting app (Teams, Zoom, Webex, FaceTime) has **no meeting URL**, so the sequence is:

1. The call stops reporting microphone input for >30s — muted, or a transient input device reconfiguration.
2. The tracker's session expires and the meeting reappears under a **new** ID.
3. `matches()` compares new-ID against old-ID with no URL to fall back on → **false, permanently**.
4. `markSourceRecovered()` can therefore never fire, so the "Meeting signal lost" warning cannot be dismissed by the meeting coming back.
5. The warning auto-dismisses 30s later and a still-running meeting is finalized and summarized.

For any non-browser meeting this teardown is **unrecoverable by construction** — once the session ID rotates there is no path back. That matches the report in #303: *"never auto-resumed even though the device was back"*, during a stretch where the microphone was demonstrably still in use by Teams.

### What changed

Fall back to app identity when — and only when — the source has no URL **and** is not a browser. Browsers stay excluded because a single browser hosts many unrelated sessions, so its bundle ID is not a meeting identity; there is a test pinning that.

This deliberately does **not** touch the 20s disappearance grace or the 30s warning window. It only restores the ability of a dedicated app source to re-match, which is what makes the existing recovery path reachable.

### Relationship to #368

Complementary rather than overlapping. #368 changes the *response* to signal loss for `.detectedPrompt` origins; this changes the *matching* that decides whether signal was lost at all. Both touch `MeetingAutoStopPolicy.swift` but different functions, and either can land first. Happy to rebase around whichever you prefer.

### Note on CI coverage

`MeetingAutoStopPolicyTests` was on the `ci_unsharded_test_suites.txt` legacy allowlist, so **none** of its coverage has been running in CI. Assigned to the `meetings` shard and removed from the allowlist; `scripts/test_ci_test_shards.sh` passes.

## Validation

Developed red-then-green rather than asserting after the fact:

- The new rotation test **fails against current `main`** with exactly the expected mismatch (`app:com.microsoft.teams2:session:1800000000` vs `...:1800000045`, `normalizedURL: nil`), and passes after the change.
- Both new negative tests — a different meeting app, and a browser-audio source with no URL — pass **before and after**, confirming the change is narrow and does not loosen browser matching.
- `swift test --filter MeetingAutoStopPolicyTests` — 20 tests pass, including all 17 pre-existing.
- `bash scripts/run_ci_test_shard.sh meetings` — **373 tests / 23 suites, all passing** (up from 351/22, since this suite now actually runs).
- `bash scripts/test_ci_test_shards.sh` — shard assignments verified.

### Still open

Whether muting is what produces these >30s gaps is app-specific and I have not verified it, so I have not claimed it as the trigger. This change is deliberately independent of the cause: it fixes the fact that *any* gap longer than the quiet window is unrecoverable for a dedicated app. If it turns out meeting apps do drop `kAudioProcessPropertyIsRunningInput` while muted, then the deeper fix is for liveness to key on the call existing rather than on microphone capture — a larger change I did not want to bundle here.

## Contribution certification

      under Muesli's
      and licenses or terms.
      resulting changes.

### Third-party materials

None.

### AI assistance

Claude Code was used to locate the root cause, implement the change, and write the tests. All changes were reviewed and tested locally, including verifying the new test fails against unpatched code.

---
<a href="https://app.blacksmith.sh/Muesli-HQ/codesmith/muesli/pr/375"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788389185&installation_model_id=422865&pr_number=375&repository=Muesli-HQ%2Fmuesli&return_to=https%3A%2F%2Fgithub.com%2FMuesli-HQ%2Fmuesli%2Fpull%2F375&signature=f483ab7cfc0db993e6dd71657f0a43158a734fbb3499475326b8c409bb3270dc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic meeting stopping when calendar events, browser tabs, or media-session identities change.
  * Improved recognition of dedicated meeting applications when meeting links are unavailable.
  * Prevented unrelated browser audio, calendar events, or different meeting applications from triggering an automatic stop.

* **Tests**
  * Added coverage for calendar-event tracking, tab changes, disappearing audio, application rotation, and unrelated meeting candidates.
  * Updated test distribution for more reliable CI execution.

### DIFF
--- a/native/MuesliNative/Sources/MuesliNativeApp/MeetingAutoStopPolicy.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/MeetingAutoStopPolicy.swift
@@ -75,6 +75,7 @@ struct MeetingAutoStopSource: Equatable {
     let suppressionID: String?
     let normalizedURL: String?
     let sourceBundleID: String?
+    let calendarEventID: String?
     let hasObservedCandidate: Bool
 
     private init(
@@ -82,12 +83,14 @@ struct MeetingAutoStopSource: Equatable {
         suppressionID: String?,
         normalizedURL: String?,
         sourceBundleID: String?,
+        calendarEventID: String?,
         hasObservedCandidate: Bool
     ) {
         self.candidateID = candidateID
         self.suppressionID = suppressionID
         self.normalizedURL = normalizedURL
         self.sourceBundleID = sourceBundleID
+        self.calendarEventID = calendarEventID
         self.hasObservedCandidate = hasObservedCandidate
     }
 
@@ -96,6 +99,7 @@ struct MeetingAutoStopSource: Equatable {
         self.suppressionID = candidate.suppressionID
         self.normalizedURL = candidate.url
         self.sourceBundleID = candidate.sourceBundleID
+        self.calendarEventID = candidate.calendarEventID
         self.hasObservedCandidate = true
     }
 
@@ -107,6 +111,7 @@ struct MeetingAutoStopSource: Equatable {
         self.suppressionID = normalized.id
         self.normalizedURL = normalized.url
         self.sourceBundleID = nil
+        self.calendarEventID = nil
         self.hasObservedCandidate = false
     }
 
@@ -119,6 +124,7 @@ struct MeetingAutoStopSource: Equatable {
             suppressionID: refinedSuppressionID,
             normalizedURL: normalizedURL ?? candidate.url,
             sourceBundleID: sourceBundleID ?? candidate.sourceBundleID,
+            calendarEventID: calendarEventID ?? candidate.calendarEventID,
             hasObservedCandidate: true
         )
     }
@@ -199,11 +205,31 @@ enum MeetingAutoStopPolicy {
             return true
         }
 
+        if matchesCalendarEventIdentity(candidate: candidate, source: source) {
+            return true
+        }
+
         if matchesDedicatedAppIdentity(candidate: candidate, source: source) {
             return true
         }
 
         return false
+    }
+
+    /// Every other identity here describes *how* a meeting is currently being
+    /// observed, and each can change while the same call continues: the room
+    /// URL is only reported while the call's browser tab is frontmost, and both
+    /// the candidate and suppression IDs are audio-session IDs that rotate once
+    /// their idle timeout lapses. A recording armed from a calendar event's
+    /// join URL then stops matching the very meeting it was started for, and
+    /// gets torn down mid-call. The calendar event is the one identity that
+    /// stays fixed, so match on it when both sides agree on the event.
+    private static func matchesCalendarEventIdentity(
+        candidate: MeetingCandidate,
+        source: MeetingAutoStopSource
+    ) -> Bool {
+        guard let sourceCalendarEventID = source.calendarEventID else { return false }
+        return candidate.calendarEventID == sourceCalendarEventID
     }
 
     /// A dedicated meeting app has no meeting URL, and `MeetingMediaSessionTracker`
--- a/native/MuesliNative/Sources/MuesliNativeApp/MeetingAutoStopPolicy.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/MeetingAutoStopPolicy.swift
@@ -224,11 +224,25 @@ enum MeetingAutoStopPolicy {
     /// join URL then stops matching the very meeting it was started for, and
     /// gets torn down mid-call. The calendar event is the one identity that
     /// stays fixed, so match on it when both sides agree on the event.
+    ///
+    /// The candidate must also show that a meeting is actually being observed.
+    /// `MeetingCandidateResolver.resolve` attributes *any* media activity to the
+    /// current calendar event once one exists, and its last fallback returns a
+    /// bare `cal:<id>` candidate with no meeting app at all. A calendar entry
+    /// that is only a reminder or a placeholder would then hold a recording open
+    /// for its whole duration on the strength of unrelated microphone use — a
+    /// dictation tool, say. Requiring attributed meeting audio or a room URL
+    /// keeps that fallback from standing in for a live call, while every calendar
+    /// branch backed by real meeting media still matches.
     private static func matchesCalendarEventIdentity(
         candidate: MeetingCandidate,
         source: MeetingAutoStopSource
     ) -> Bool {
         guard let sourceCalendarEventID = source.calendarEventID else { return false }
+        guard candidate.evidence.contains(.audioInputProcess)
+            || candidate.evidence.contains(.browserURL) else {
+            return false
+        }
         return candidate.calendarEventID == sourceCalendarEventID
     }
 
--- a/native/MuesliNative/Sources/MuesliNativeApp/MeetingAutoStopPolicy.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/MeetingAutoStopPolicy.swift
@@ -199,6 +199,32 @@ enum MeetingAutoStopPolicy {
             return true
         }
 
+        if matchesDedicatedAppIdentity(candidate: candidate, source: source) {
+            return true
+        }
+
         return false
+    }
+
+    /// A dedicated meeting app has no meeting URL, and `MeetingMediaSessionTracker`
+    /// mints a new session ID once its quiet window lapses. A call that briefly
+    /// stops reporting microphone input — being muted, or a transient input device
+    /// reconfiguration — therefore reappears under an ID that can never match the
+    /// armed source again, so the still-running meeting gets torn down and cannot
+    /// recover. Fall back to the app identity for that case.
+    ///
+    /// Browsers are excluded: a single browser hosts many unrelated sessions, so
+    /// its bundle ID is not a meeting identity.
+    private static func matchesDedicatedAppIdentity(
+        candidate: MeetingCandidate,
+        source: MeetingAutoStopSource
+    ) -> Bool {
+        guard source.normalizedURL == nil,
+              candidate.url == nil,
+              let sourceBundleID = source.sourceBundleID,
+              MeetingCandidateResolver.browserApps[sourceBundleID] == nil else {
+            return false
+        }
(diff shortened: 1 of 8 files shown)


jamesagarside:
Flagging a trade-off I under-stated in the description, rather than leaving it to be found in review.

Falling back to app identity makes the match **broader**: any candidate from the same dedicated app now matches the armed source. The failure mode that introduces is the mirror of the one it fixes — if a meeting app keeps its microphone and speaker streams open *after* a call ends, the candidate keeps matching, `lastSeenAt` keeps refreshing, and the recording may never auto-stop. Related: leaving one call and joining another in the same app now continues the existing recording rather than treating it as a new meeting.

I still think that is the right direction, for three reasons:

1. It only affects auto-started recordings — `.manual` never arms auto-stop at all.
2. Explicit stop is always available, so the cost is a recording that runs long rather than one that is lost. The current failure loses transcript from a live meeting, which is unrecoverable.
3. The existing behaviour is not a conservative default — it is unrecoverable by construction once the session ID rotates, which is the part I would argue is the actual bug.

If you would rather bound it, the narrow version is to require that the app also still be running (a `sourcePID` liveness check) before the identity fallback applies. Happy to add that if you prefer a tighter guard.

On process: I am conscious this touches the same file as #368. If you would rather have one change here than two, I am glad to close this and fold the matching fix into that PR, or rebase on top of it once it lands — whichever keeps your review queue smaller. No attachment to it being a separate PR.

jamesagarside:
I've pushed a second commit here, because a real-world failure turned out to be the same defect with a different identity — and I'd rather widen this PR than open a third one.

### What happened

A "Join & Record" recording was torn down 44 seconds in, with the "Meeting signal lost" prompt, while the call was still going. It isn't a one-off: the same user has 9 recordings under two minutes going back several weeks.

### Why

`matches()` compares the armed source against the live candidate using `candidateID`, `suppressionID` and `normalizedURL`. Every one of those describes *how* the meeting is currently being observed, and each can change while the same call continues:

- `normalizedURL` is only present while the call's browser tab is frontmost. `BrowserMeetingActivityCollector` reads `kAXDocumentAttribute` per window and the ScriptingBridge `activeTab`, so a meeting in a background tab reads as `.noMeeting` — which also *removes* the TTL cache entry rather than letting it age out.
- `candidateID` and `suppressionID` are audio-session IDs from `appAudioSessionID(forBundleID:prefix:now:)`, which mints a fresh ID once `appAudioSessionIdleTimeout` (10s) lapses without a refresh. The `browserMeeting.isFocused` branch of `resolve()` never calls it, so a focused tab with no active input process — muted, for instance — lets the session expire.

When any of that happens, `MeetingCandidateResolver` *still resolves the meeting*, through the calendar-keyed fallback (`id: "cal:\(calendarEvent.id)"`). The candidate is right there, attributed to the very event the recording was started for. `matches()` just can't see it, so the grace period elapses and the recording stops.

"Join & Record" is hit hardest because it arms auto-stop from `MeetingAutoStopSource(meetingURL:)`, so the room URL is the only identity the source has to begin with.

### The change

Carry the active calendar event ID on `MeetingCandidate`, learn it into `MeetingAutoStopSource` via the existing `refined(with:)` step, and match on it.

This is tighter than the app-identity fallback in the first commit: it matches only when both sides agree on the *same calendar event*, so it can't bleed into an unrelated meeting, and a source with no calendar event behaves exactly as before. It's also robust to all three triggers above rather than any one of them, which matters because I could not prove from the user's data which one fired — only that every identity the matcher relies on is derived from the observation rather than the meeting.

Verified by mutation: dropping the new match arm fails exactly the two behavioural tests and leaves the three negative-control tests passing.

### Known gap

A browser meeting with **no** calendar event still has no stable identity, so backgrounding its tab can still tear it down. Fixing that means either treating a background tab as a live meeting or keying liveness on the browser still holding the mic — both change what counts as a live meeting for auto-*start* too, which is a much bigger blast radius than I'd want in this PR. Happy to follow up separately if you think that's worth doing.

Still glad to reshape this however suits you — including splitting the two commits back apart if you'd rather review them independently.

jamesagarside:
Rebased onto current `main`, and pushed a third commit fixing a defect in this
PR's own second commit.

**`matchesCalendarEventIdentity` was unreachable.**
`MeetingMediaSessionTracker.stabilize` rebuilds the candidate to swap in the
stable session identity, and it did not carry `calendarEventID` across. The
field defaults to nil, so it was dropped silently — no compiler error, no test
failure.

`stabilize` sits between `resolve` and the auto-stop matcher
(`MeetingMonitor.swift`), so the stripped candidate is what reaches
`MeetingAutoStopPolicy.matches`. The match arm added in `0204e01f` could
therefore never fire for a media-backed candidate — and on the calendar path
every candidate is media-backed, because `resolve` only reaches the calendar
fallbacks once `hasMediaActivity` holds. The commit was inert.

The session now carries the calendar event the same way it already carries `url`
and `meetingTitle`: prefer the newest observation, fall back to what the session
already knew.

**Why CI did not catch it.** `MeetingMediaSessionTrackerTests` was on
`scripts/ci_unsharded_test_suites.txt`, so it never ran in required CI. Moved
into the meetings shard, alongside the two suites this PR already moved.

Added a regression test that asserts the session identity is rewritten while the
calendar attribution survives — it fails against the previous commit with
`result?.calendarEventID → nil`.

Full suite: 1546 tests across 150 suites, passing.

Separately, field testing on 2026-08-08 found the muting hypothesis in the
original description does not hold: a two-minute Zoom mute during an active
recording left the audio process at `IsRunningInput=1 / IsRunningOutput=1`
throughout, and the recording survived. Session-ID rotation from an idle-timeout
lapse is still a real failure mode and this PR still addresses it, but muting is
not the trigger that produces it.

jamesagarside:
Good catch on the untested path. Added in `927bd893`.

The existing test only exercised session *creation*. Every candidate after the
first takes the refresh path in `updateSession`, where the stored event id is
updated rather than set, and that had no coverage at all.

The new test drives three stabilize calls against one session: a newer event id
replaces the stored one, and a later candidate carrying no event id leaves it
intact. The second half is the one with teeth — a live meeting emits candidates
continuously and the cached calendar event drops off whenever it lapses
mid-call, so treating absence as a new answer would clear the id and take
`matchesCalendarEventIdentity` dark for the rest of the recording. Same failure
as dropping it on rebuild, just arriving later in the call.

Mutation-verified: clearing on absence fails the preserve assertion, removing
the update line fails both. Neither mutation is caught by the pre-existing
creation-path test, which is the gap you flagged.

jamesagarside:
Pushed `8e86b18d` after a live reproduction last night with per-process
CoreAudio sampling running throughout. It exposed a gap that made the
dedicated-app fix in `9a11dd1b` unreachable for the entry point most
recordings actually use.

```
22:01:03  Zoom   in=0 out=0 -> in=1 out=1   call joined
22:04:31  Muesli in=0       -> in=1         recording starts
22:05:55  Muesli in=1       -> in=0         torn down 84s in
23:07:42  Zoom   in=1 out=1 -> in=0 out=0   call actually ends
```

Zoom held `in=1 out=1` continuously across the teardown — not one transition
between 22:01:03 and 23:07:42. The call never stopped qualifying, so this is
not signal loss. The candidate was there; `matches()` stopped recognising it.

The recording was started from a "Join & Record" notification, which arms via
`MeetingAutoStopSource(meetingURL:)`. That leaves the source with URL-derived
identity only — no bundle id, no calendar event. A dedicated meeting app's
candidates carry no URL, so the URL arm cannot match them, and:

- `matchesDedicatedAppIdentity` opened with `guard source.normalizedURL == nil`,
  and `refined` never clears `normalizedURL` (`normalizedURL ?? candidate.url`).
  So for every Join & Record recording that arm was disabled for the life of
  the recording.
- The source could not rescue itself by refining, either. It only refines after
  a successful match, and the only candidates that match a URL are browser
  ones — so it could never learn the meeting app's bundle id.

Dropping the guard alone would therefore have been inert: the arm also needs
`source.sourceBundleID`, which a URL-armed source can only ever acquire as a
*browser* bundle id, which the browser exclusion then correctly rejects. So
both halves are here:

1. Carry the calendar event id into the source when arming from a join link.
   Both sides derive it from the same EventKit `eventIdentifier`, so it is an
   identity they can agree on from the first candidate rather than waiting on a
   refinement that never arrives. This is the half that reaches production.
2. Key the dedicated-app arm off the candidate having no URL rather than the
   source. The browser exclusion still carries the original intent.

Mutation-verified separately: dropping the event id at arm time fails only the
join-link calendar test; restoring the source-URL guard fails only the rotation
test. Negative controls — a different calendar event, and an unrelated browser
session on a source that learned a browser — keep passing. Full suite 1558/1558.

One thing I cannot show from here: which of the two arming paths produced last
night's teardown. Reading Muesli's own os_log needs Full Disk Access this
account does not have, so I can distinguish the code paths but not confirm
which one ran. The gap above is a property of the code and holds regardless.

PR #376, opened 2026-08-04
Add system theme matching option

slenderdeveloperman:
## Summary

**This PR pivoted mid-review — see "Design pivot" below before reviewing.** The current diff replaces `darkMode: Bool` with a three-case `ThemeMode` (`light` / `dark` / `system`):

- `.system` maps to `.preferredColorScheme(nil)`, which lets SwiftUI defer to whatever macOS's current appearance actually is — including the OS's own "Automatic" mode (System Settings → Appearance) if the user has that enabled, which already switches dark/light based on sunrise/sunset using Location Services the OS manages.
- Sidebar sun/moon/circle-half segmented control and the Settings → Appearance pane both now just select `ThemeMode` directly.
- Backward compatible: existing installs have a persisted `dark_mode` boolean in `config.json`. Decoding now reads `theme_mode` first and falls back to migrating the legacy `dark_mode` key (`true → .dark`, `false → .light`) so nobody's existing preference resets on upgrade.

## Design pivot

The first two commits on this branch implemented automatic dark/light switching driven by a custom NOAA-formula sunrise/sunset calculator (`SolarCalculator.swift`) with manually-entered latitude/longitude. Greptile's review caught three real P1s in that version (UTC-day boundary bug in the solar math, a silently-nonfunctional auto-mode-without-coordinates state, and unclamped lat/long input), which were fixed in a follow-up commit — but on reflection the whole approach was solving a problem macOS already solves:

- The original code **forced** `.preferredColorScheme(.dark/.light)` unconditionally — Muesli could never follow the OS's own appearance setting at all. That was the actual gap.
- macOS's built-in "Automatic" appearance mode already does sunrise/sunset-based switching, using real device location (via Location Services) rather than a manually-typed coordinate pair, with none of the maintenance burden of hand-rolled solar position math.

So this final version deletes `SolarCalculator.swift` and its tests, the `autoThemeByDaylight`/`themeLatitude`/`themeLongitude` config fields, and the lifecycle-observer/fallback-timer plumbing in `MuesliController`, in favor of the much smaller `ThemeMode` enum. Net diff across the full branch history is now additive by only ~80 lines instead of the ~260 the sunrise/sunset version added.

## Test plan
- [x] `swift test --package-path native/MuesliNative` — full suite passes (1,483 tests, 145 suites). Two unrelated pre-existing flaky concurrency tests (`DiarizerRuntimePolicyTests`, `StreamingVadControllerTests`) were confirmed to pass in isolation — not caused by this change.
- [x] New `ConfigStoreTests` cases cover the legacy `dark_mode` → `themeMode` migration, including the case where both keys are present (new key wins).
- [x] `swift build --package-path native/MuesliNative` — builds cleanly.
- [ ] Manual verification in `MuesliDev.app` (not yet run in this environment — no macOS GUI session available here).

## AI-assisted contribution disclosure
This change was written with assistance from Claude Code (Anthropic), including the mid-review pivot away from the original sunrise/sunset design. I reviewed the diff, the Greptile findings on the earlier version, and the migration logic before pushing. No third-party code, data, or model output beyond the assistant's generated Swift was incorporated.

## Summary by CodeRabbit

* **New Features**
  * Added Light, Dark, and System appearance modes.
  * System mode follows the macOS appearance setting, including automatic changes throughout the day.
  * Updated appearance controls to clearly indicate the active selection.

* **Bug Fixes**
  * Existing dark-mode preferences are automatically preserved when upgrading.
  * Appearance settings now persist correctly, including System mode.


slenderdeveloperman:
Following up on @greptile-apps's three P1 findings on the earlier sunrise/sunset version (UTC-day anchoring bug in `SolarCalculator`, auto-mode-without-coordinates no-op, unclamped lat/long input):

All three were fixed in [149a43e8](https://github.com/Muesli-HQ/muesli/commit/149a43e8), but on reflection the underlying design was the wrong shape — it duplicated macOS's own "Automatic" appearance mode (which already does sunrise/sunset switching via real Location Services) with hand-rolled solar math and manual coordinate entry.

[f94b6445](https://github.com/Muesli-HQ/muesli/commit/f94b6445) replaces the whole approach: `SolarCalculator` and the coordinate-based auto mode are gone, replaced by a plain three-way `ThemeMode` (Light/Dark/System) where "System" just passes `nil` to `.preferredColorScheme` and lets the OS (including its own Automatic mode, if the user enables it) drive things. See the updated PR description for the full rationale — all three original findings are moot under this design since there's no more custom solar math or coordinate input to have bugs in.

PR #388, opened 2026-08-08
Keep dictation capture graph persistent; eliminate idle route-change churn

pHequals7:
## Why

Two problems share one root cause — Muesli tore down and rebuilt the dictation capture graph on every audio route change and after every dictation:

1. **Idle coreaudiod churn**: with no meeting/dictation active, each AirPods connect/disconnect caused duplicate AudioQueue teardown/rebuild cycles and daemon CPU bursts (measured up to 175–193% transient, ~2× baseline median in A/B), with all work attributed to an idle Muesli — the only Muesli-owned CoreAudio activity matching field reports of daemon load while idle.
2. **Hotkey latency**: rebuilding the graph on route events and after every dictation put graph construction back into the hot path.

Reference architecture validated on this machine: Other apps keep pre-created daemon-side IO contexts alive indefinitely and only starts/stops IO per dictation (~65ms hotkey→mic, zero work on route notifications, verified via coreaudiod unified logs across dictations and an AirPods case→out transition). Our AudioQueueInputRecorder already had the right two-phase shape — the teardown was policy, not necessity.

## What changes (persistent-graph lifecycle)

- **`AppScopedDictationRecorder.stop()`** no longer cancels the child recorder. `AudioQueueInputRecorder.stop()` already retains the prepared queue, so the next dictation is `AudioQueueStart` on the existing graph — no rebuild.
- **`RouteAwareDictationRecorder`** keeps both slot graphs warm across `stop()` and slot switches (previously cancelled the inactive slot).
- **`DictationAudioSessionManager`** treats idle route-change refreshes as telemetry-only — no coolDown, no re-warm. The capture graph binds to a fixed input device and is indifferent to output-route churn.
- **Default-route reuse is safe without any rebind logic**: a queue prepared without an explicit device reports `kAudioQueueProperty_CurrentDevice = AQDefaultDevice` and follows the system default input at each `AudioQueueStart` — verified empirically (probe: same queue, default input flipped built-in→BlackHole between starts; coreaudiod logs show the restart capturing from BlackHole). Explicit-device queues bind by UID; a disappeared device fails at `AudioQueueStart` and the existing failure path disposes + rebuilds.

**Rebuilds now happen only on proven failure** (start error → existing fallback recorder, no-audio timeout, explicit cancel) **or device-selection change.** Startup warmup and arm-time activation are unchanged.

## Validation

- 60/60 focused dictation tests updated to the new lifecycle contract; full suite 1524/1524 green.
- Live transition matrix on a dev lane (macOS 26.5.2): repeat dictations reuse the warm graph (`audio_queue_prepare_reused`, no `new_input` outside selection changes); AirPods connect with warm graph → zero churn, 32ms start; explicit AirPods-mic selection → one rebuild on selection change (expected), BT HFP start tax ~500-600ms (daemon-side, not graph); AirPods cased with explicit selection → graceful resolve to warm built-in graph, 78ms start; reconnect → one cheap rebuild, then reuse.
- Mic activation now ~32-78ms on built-in routes (parity with Willow's measured ~65ms).

## Contribution Certification

## Third-Party Materials

None — no third-party code, libraries, or assets added. (Willow Voice was observed externally via its daemon telemetry as a reference architecture; no code or assets were taken from it.)

## AI Assistance

This change was developed with AI assistance (OpenAI Codex): comparative architecture analysis and implementation. CoreAudio lifecycle claims were validated by direct local experiment on macOS 26.5.2.

### DIFF
--- a/native/MuesliNative/Sources/MuesliNativeApp/AppScopedDictationRecorder.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/AppScopedDictationRecorder.swift
@@ -280,9 +280,12 @@ final class AppScopedDictationRecorder: DictationAudioRecording {
         lifecycleGeneration &+= 1
         lock.unlock()
         return recorderQueue.sync {
-            let url = recorder.stop()
-            recorder.cancel()
-            return url
+            // Stop IO and finalize the WAV, but keep the prepared capture
+            // graph alive: AudioQueueInputRecorder.stop() retains the queue,
+            // so the next start() is AudioQueueStart on the existing graph
+            // instead of a full rebuild. The graph is only torn down on
+            // cancel() (proven failure / explicit invalidation).
+            recorder.stop()
         }
     }
 
--- a/native/MuesliNative/Sources/MuesliNativeApp/AudioQueueInputRecorder.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/AudioQueueInputRecorder.swift
@@ -260,7 +260,8 @@ final class AudioQueueInputRecorder: StreamingDictationRecording, StreamingDicta
         emitLatency("audio_queue_prepare_end")
     }
 
-    private func applyPreferredInputDeviceID(_ deviceID: AudioObjectID, to queue: AudioQueueRef) throws {        emitLatency("audio_queue_device_uid_lookup_begin")
+    private func applyPreferredInputDeviceID(_ deviceID: AudioObjectID, to queue: AudioQueueRef) throws {
+        emitLatency("audio_queue_device_uid_lookup_begin")
         guard var deviceUID = Self.deviceUID(for: deviceID) as CFString? else {
             throw Self.runtimeError(code: 6, message: "Could not resolve device UID for \(deviceID)")
         }
--- a/native/MuesliNative/Sources/MuesliNativeApp/AudioQueueInputRecorder.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/AudioQueueInputRecorder.swift
@@ -186,6 +186,12 @@ final class AudioQueueInputRecorder: StreamingDictationRecording, StreamingDicta
     }
 
     private func prepareLocked() throws {
+        // Reuse is safe for the default route too: a queue prepared without an
+        // explicit device reports kAudioQueueProperty_CurrentDevice =
+        // "AQDefaultDevice", i.e. it follows the system default input at start
+        // time. Explicit-device queues are bound to a fixed UID; if that device
+        // disappears, AudioQueueStart fails and the caller's failure path
+        // disposes + rebuilds. Neither case needs eager rebuild here.
         if isPrepared, preparedInputDeviceID == preferredInputDeviceID {
             emitLatency("audio_queue_prepare_reused")
             return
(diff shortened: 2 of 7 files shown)


pHequals7:
Evidence against the P1 ("Default-input queue retains the prior microphone"):

The claim assumes an AudioQueue created without an explicit device binds to the default input as of creation. Empirically it does not: such a queue reports `kAudioQueueProperty_CurrentDevice = AQDefaultDevice` (a sentinel meaning 'follow the system default'), and the binding is resolved at each `AudioQueueStart`, not at creation.

Direct experiment on macOS 26.5.2 (probe: create input queue with no explicit device → start with built-in mic as default → stop → flip default input to BlackHole 2ch → restart the SAME queue object):

Daemon-side ground truth (coreaudiod unified log):

```
12:06:02.015  IOWorkLoopInit: ctx 130952 BuiltInMicrophoneDevice  ← phase 1 (default = built-in)
12:06:02.964  IOWorkLoopInit: ctx 130948 BlackHole2ch_UID         ← phase 2 (same queue, default flipped)
12:07:52.106  IOWorkLoopInit: ctx 131192 BuiltInMicrophoneDevice  ← repeat run, phase 1
12:07:53.073  IOWorkLoopInit: ctx 131188 BlackHole2ch_UID         ← repeat run, phase 2
```

The retained default-input queue captured from the new default device on restart — no staleness, no capture from a disconnected microphone. (Mid-recording default changes are out of scope for both old and new behavior: route refresh has always been idle-gated, and a running queue keeps its device until stop in either design.)

The explicit-device path (user-selected mic) is unaffected by this concern: those queues bind by device UID, and a disappeared device fails at `AudioQueueStart`, which the existing fallback path handles by disposal + rebuild.

So the reuse guard is correct as written; the `AQDefaultDevice` sentinel semantics make the default-route queue self-updating across idle default-input changes.

pHequals7:
@greptile please re-review — the P1 has been addressed with empirical daemon-log evidence in the comment above (no code change needed; the AQDefaultDevice sentinel follows the default input at each AudioQueueStart).

pHequals7:
@claude — all three points addressed:

1. **Dead code**: removed `inactiveRecorderLocked()` and `inactiveRecorder(for:)` in 265c53a5.
2. **Formatting**: fixed the merged signature line at AudioQueueInputRecorder.swift:263 in the same commit (was a stray edit artifact, good catch).
3. **Risky-route cooldown removal — confirmed intentional, and now live-validated.** The design rationale: an idle, prepared AudioQueue bound to a fixed input device costs nothing while not running (no IO, no indicator, no daemon work) — so there is no safety benefit to tearing it down on a 'risky' route; the risk framing came from an era when warm graphs were re-created eagerly on route flips, which is exactly the churn this PR removes. Live validation on a dev lane (macOS 26.5.2) covering precisely this transition: AirPods connect with a warm graph → zero queue churn in coreaudiod logs, next dictation reused the warm graph and started in 32ms; AirPods cased with an explicit AirPods-mic selection → policy resolved to the warm built-in-mic graph and dictated successfully in 78ms; reconnect → one cheap rebuild at arm (device ID changed), then reuse. Telemetry: `audio_queue_prepare_reused` on every arm/start, zero `new_input` outside selection changes.

On the CodeRabbit item you flagged: agreed — its suggested test names reference identifiers from an intermediate revision that no longer exist; no action taken there.

pHequals7:
Addressing Devin's 🔴 ("Dictation can keep using the old microphone after the user changes the system default input"):

**The AudioQueue half is disproven empirically** — see the daemon-log evidence comment above (probe: same queue object, default input flipped built-in → BlackHole between starts; coreaudiod shows the restart capturing from BlackHole). A queue reporting `AQDefaultDevice` resolves the default input at each `AudioQueueStart`, not at creation, so no rebuild path is needed for the default slot: it is not pinned to the old microphone. The scenario 'user changes default mic while idle → next dictation uses old mic' does not occur — that is precisely what the probe measured.

Mid-recording default changes are behavior-parity with the old code (route refresh was already idle-gated; a running queue keeps its device in both designs).

**The StreamingMicRecorder fallback half**: the fallback engine is never kept warm across dictations — it is prepared fresh inside the failure path (`prepareFallbackLocked` runs only when the primary fails, after `primary.cancel()`), so there is no long-lived stale fallback graph to reuse. And a stopped AVAudioEngine's inputNode tracks the *current* default input at start; Muesli's own recovery history (config-change notifications firing under running engines) exists precisely because engines follow the route. `recoversFromInputConfigurationChanges == false` disables mid-stream auto-recovery, not default-tracking at start.

**The description catch is fair**: the PR body claimed 'lazily rebinds at next use' — that logic existed in an intermediate commit and was removed once the AQDefaultDevice follow-at-start semantics were verified. The body now describes the actual mechanism. Thanks for the careful read.

Issue #391, opened 2026-08-10
Resuming a finished meeting silently disarms auto-stop — what should resume inherit?


jamesagarside:
## Background

`resumeFinishedMeeting` re-arms the auto-stop tracker with a hardcoded
`MeetingRecordingStartOrigin.manual` (`MuesliController.swift`). Because
`.manual` has `enablesMeetingAutoStop == false`, resuming a meeting silently
discards auto-stop: `signalLossSource` returns `nil` and `signalLossResponse`
becomes `.none`.

A recording armed from a detection prompt, calendar auto-record, scheduled
prompt, or Join & Record therefore comes back unstoppable and runs until the
user stops it by hand.

Observed in a live session on 2026-08-08:

```
10:13:48  Zoom call ends — audio process drops to in=0 out=0
10:14:22  auto-stop fires correctly, meeting finalised (387s, status 'completed')
10:14:35  user clicks Resume — status forced back to 'recording'
10:14:36  mic reacquired; recording continues for a further ~15 minutes
```

Auto-stop was not broken. Resume dropped it.

A fix that inherits the original origin is proposed in the linked PR. This
issue is for the design question that fix does not settle.

## The design question

Resume reopens the *same* meeting row. Should the reopened recording behave
like the meeting it is, or like a fresh manual recording?

### Option A — inherit the original origin (proposed in the PR)

Record how each meeting was started and re-arm resume with that origin.

- Matches the mental model: it is still the same meeting, so it should still
  behave like that meeting.
- The reason a user resumes is usually that they disagreed with the stop. Coming
  back with auto-stop disarmed is the opposite of what they asked for.
- **Limitation:** the record is in-memory, so a meeting started before the
  current app launch has no known origin and falls back to `.manual`. Resume is
  deliberately age-independent (`MeetingResumePolicy` notes users may resume
  much older meetings), so this gap is not rare.

### Option B — persist the start origin on the meeting row

Add a column to `meetings` and carry it through the store and CloudKit sync.

- Removes Option A's limitation entirely; resume behaves the same on day 30 as
  on minute 2.
- Costs a schema migration and a sync field for something no other feature
  currently reads.
- Question: should the origin be part of the meeting's durable identity at all,
  or is it start-time scaffolding that legitimately expires?

### Option C — re-arm from current conditions rather than history

Ignore the original origin. On resume, arm auto-stop if a live meeting candidate
is currently observed, and leave it disarmed otherwise.

- Needs no new state, persisted or in-memory.
[...]

PR #392, opened 2026-08-10
fix: keep auto-stop armed when a finished meeting is resumed

jamesagarside:
## Summary

`resumeFinishedMeeting` armed the auto-stop tracker with a hardcoded
`MeetingRecordingStartOrigin.manual`. That origin has
`enablesMeetingAutoStop == false`, so `signalLossSource` returns `nil` and
`signalLossResponse` is `.none` — resuming silently discarded auto-stop.

A recording armed from a detection prompt, calendar auto-record, scheduled
prompt, or Join & Record therefore came back unstoppable and ran until stopped
by hand.

Observed in a live session on 2026-08-08:

```
10:13:48  Zoom call ends — audio process drops to in=0 out=0
10:14:22  auto-stop fires correctly, meeting finalised (387s, status 'completed')
10:14:35  Resume clicked — status forced back to 'recording'
10:14:36  mic reacquired; recording continues for a further ~15 minutes
```

Auto-stop was working. Resume dropped it.

A second reproduction on 2026-08-10, this time with per-process CoreAudio
sampling running throughout, so the call's own state is recorded rather than
inferred. Recording started from a "Join & Record" notification:

```
22:01:03  Zoom   in=0 out=0 -> in=1 out=1   call joined
22:04:31  Muesli in=0       -> in=1         recording starts
22:05:55  Muesli in=1       -> in=0         torn down 84s in, while Zoom held in=1 out=1
22:06:01  Muesli in=0       -> in=1         resumed by hand, call still live
23:07:42  Zoom   in=1 out=1 -> in=0 out=0   call actually ends
23:20:09  ...                               recording finally stopped, 12.5 min later
```

Meeting row: 4534s, 17800 words, `meeting_status 'completed'`.

Zoom held `in=1 out=1` continuously from 22:01:03 to 23:07:42 — not one
transition. The teardown at 84s happened while the call was plainly still
there, and the 12.5 minutes at the end were recorded into an empty room.

This is the resume-during-a-live-call case, which is the one this PR fixes: at
22:06:01 the call was still running, so `recentMeetingAutoStopSource()` would
have found a candidate inside its 15s window and re-armed. Whether the
re-armed recording then survived to 23:07:42 is a separate question that
belongs to #375 — the source would be candidate-derived, so it depends on the
dedicated-app arm holding across session rotation. The two changes compose;
neither alone covers this timeline.

Resume reopens the *same* meeting row, so it now inherits the origin that row
was started with. `MeetingResumePolicy.resumedStartOrigin` makes the decision
explicit and testable; `MuesliController` records each start origin against its
meeting id and reads it back on resume.

### Scope — what this does and does not fix

Restoring the origin selects the intended *response*, but arming also needs a
*source*. `armMeetingAutoStop` sets the response to `.none` when the source is
nil, and `MeetingAutoStopTracker.isArmed` is `source != nil`. So the two resume
cases differ:

- **Resume while the call is still live** — the common case, where the user
  resumes because they disagreed with the stop. `handleMeetingActivityCandidate`
  keeps repopulating `latestMeetingActivityCandidate` while unarmed and not
  recording, so `recentMeetingAutoStopSource()` finds a source inside its 15s
  window and auto-stop re-arms. **This PR fixes this case.**
- **Resume after the call has ended** — no candidate is observed, the source is
  nil, and the recording stays unarmed until stopped by hand. **This PR does not
  change this case.**

The 2026-08-08 incident above was the second case: the call ended at 10:13:48
and resume happened at 10:14:35, after it was over. This PR would not have
prevented that recording. Credit to Greptile's review for catching that the
original description of this PR overstated its reach.

Closing the second case needs a mechanism that does not exist yet — auto-stop
has no source to track when no call is in progress, so it would mean stopping on
the *absence* of a candidate after a grace window rather than on the
disappearance of a known one. That is a behaviour change with its own failure
modes and belongs in #391, not smuggled into this PR.

A second limitation: the origin record is in-memory, so a meeting started before
the current app launch has no known origin and falls back to `.manual`. Resume
is deliberately age-independent, so this gap is not rare.

## Validation

- `swift test --package-path native/MuesliNative` — 1546 tests in 150 suites,
  passing. One run flagged `ConfigStore`
  ("saved config uses owner-only file permissions"); it passes in isolation both
  with and without this change and did not recur on re-run, so it looks like
  parallel-execution interference rather than a regression from this change.
- Mutation-verified: forcing `resumedStartOrigin` back to `.manual` fails
  "resuming an auto-armed meeting keeps auto-stop armed" and
  "a resumed meeting inherits the origin it was started with", while the manual
  and unknown-origin negative controls keep passing.
- `MeetingResumePolicyTests` is already assigned to a real CI shard
  (`scripts/run_ci_test_shard.sh`), so the new tests run in required CI.

Not covered by tests: the wiring in `resumeFinishedMeeting` itself.
`MuesliController` is a large `@MainActor` class with no existing unit-test
seam, so the tests cover the policy decision rather than the call site. The call
site was verified by reading the diff, not by an automated test.

## Contribution certification

      under Muesli's
      and licenses or terms.
      resulting changes.

### Third-party materials

None.

### AI assistance

Claude Code (Anthropic) was used materially: it diagnosed the defect from live
CoreAudio process sampling and the local meeting database, wrote the change and
its tests, and ran the test suite and mutation check. The change and the test
results were reviewed before submission.

### DIFF
--- a/native/MuesliNative/Sources/MuesliNativeApp/MeetingResumePolicy.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/MeetingResumePolicy.swift
@@ -31,4 +31,22 @@ enum MeetingResumePolicy {
     static func hasNewTranscriptContent(prior: String, new: String) -> Bool {
         combinedResumeTranscript(prior: prior, new: new) != prior
     }
+
+    /// The start origin a resumed meeting should be re-armed with.
+    ///
+    /// Resume reopens the *same* meeting row, so it should keep behaving like the
+    /// meeting it is. Hardcoding `.manual` here silently discarded auto-stop:
+    /// a recording armed from a detection prompt or a calendar event became
+    /// unstoppable the moment the user resumed it, and ran until stopped by hand.
+    ///
+    /// `recordedOrigin` is nil when the original start is not known — the meeting
+    /// was started in an earlier app session, so nothing in memory records how it
+    /// began. Falling back to `.manual` there preserves the previous behaviour for
+    /// exactly the case where inheriting is not possible, and never re-arms a
+    /// recording against a meeting this process never observed starting.
+    static func resumedStartOrigin(
+        recordedOrigin: MeetingRecordingStartOrigin?
+    ) -> MeetingRecordingStartOrigin {
+        recordedOrigin ?? .manual
+    }
 }
--- a/native/MuesliNative/Sources/MuesliNativeApp/MuesliController.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/MuesliController.swift
@@ -431,6 +431,12 @@ final class MuesliController: NSObject {
     /// Present only while a resume is in flight; consumed at stop to merge old + new
     /// transcript, and cleared on success or restored-on-failure.
     private var pendingResumePriorTranscript: [Int64: String] = [:]
+    /// How each meeting recorded in this app session was started, keyed by meeting id.
+    /// Resuming a finished meeting reads this so the reopened recording keeps the
+    /// auto-stop behaviour it was armed with instead of degrading to `.manual`.
+    /// In-memory only: meetings started before the current launch are absent, and
+    /// `MeetingResumePolicy.resumedStartOrigin` falls back to `.manual` for those.
+    private var meetingStartOrigins: [Int64: MeetingRecordingStartOrigin] = [:]
     private var iCloudSyncTask: Task<Void, Never>?
     private var ckSyncEngine: MuesliCKSyncEngine?
     private var ckSyncEngineCancellationTask: Task<Void, Never>?
@@ -4845,6 +4851,7 @@ final class MuesliController: NSObject {
             )
             activeMeetingID = meetingID
             activeMeetingAudioWarning = nil
+            meetingStartOrigins[meetingID] = startOrigin
             syncAppState()
             if openDocument {
                 showMeetingDocument(id: meetingID)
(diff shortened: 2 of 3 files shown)


jamesagarside:
Both reviews land on the same real gap, and it is the one the "Scope" section of
the description already calls out: restoring the origin selects the *response*,
but arming also needs a *source*, and resuming after the call has ended has no
candidate to build one from. Greptile caught that in the first round and the
description was corrected accordingly.

One correction on the suggested fix, though. CodeRabbit proposes storing the
effective `MeetingAutoStopSource` alongside the origin and passing it as
`explicitSource` on resume. That does not close the case it names.

`MeetingAutoStopTracker.observe` returns `false` whenever `lastSeenAt` is nil:

```swift
guard let lastSeenAt else {
    return false
}

return now.timeIntervalSince(lastSeenAt) >= gracePeriod
```

`lastSeenAt` is only set by `markRecordingStarted(now:)`, which is itself a
no-op unless `observedBeforeRecordingStarted` is true — and that flag is only
set by `observeBeforeRecordingStarted` matching a *live* candidate before the
recording starts. In the ended-call case there is no live candidate by
construction, so a restored source would arm the tracker (`isArmed` is
`source != nil`) without ever starting the disappearance clock. The recording
would still run until stopped by hand.

`MeetingAutoStopPolicyTests` already pins this behaviour:
`trackerWaitsForRecordingTimeObservationBeforeAutoStopping` arms a valid source,
observes a nil candidate 25s later against a 20s grace period, and asserts
`!shouldStop` and `lastSeenAt == nil`. That test is the ended-call resume
scenario exactly.

So persisting the source is necessary but not sufficient. Closing the case needs
stopping on the *absence* of a candidate rather than the disappearance of a
known one — a different mechanism with its own failure modes, and one that pulls
against what #375 exists to prevent. It is written up in #391 for a maintainer
decision rather than smuggled in here.

On the P1 severity: the description of current behaviour is accurate, but this
is a pre-existing gap that the PR does not widen. Before this change the origin
was hardcoded to `.manual`, whose `enablesMeetingAutoStop` is false, so a
resumed recording was unarmed in *both* cases. After it, resume during a live
call re-arms and resume after the call ends behaves as it did before. Strict
improvement, no regression.

Happy to take the other route if a maintainer prefers it — persisting the source
is a reasonable thing to want for its own sake, it just needs the absence-based
stop to go with it to have the claimed effect.

PR #394, opened 2026-08-10
feat: optionally require a join link for calendar meeting detection

jamesagarside:
## Summary

Meeting detection accepts any timed, non-all-day calendar entry inside its
window. `selectCurrentOrNearbyCachedCalendarEvent` filters on three things only:

```swift
!event.isAllDay && event.endDate > now && event.startDate < searchEnd
```

So a reminder, a placeholder or a focus block is an ordinary meeting as far as
detection is concerned. That matters because once an event is in play,
`MeetingCandidateResolver.resolve` attributes *any* media activity to it, and its
last calendar fallback returns a bare `cal:<id>` candidate with no meeting app at
all — a dictation tool holding the microphone is enough to produce a candidate
stamped with that event.

`UnifiedCalendarEvent` already carries `meetingURL`; selection simply never
consulted it. The new `require_calendar_meeting_link` setting makes selection
skip events without one, with a toggle in Settings → Meetings.

### Why it is off by default

Link-less entries are still real meetings for anyone who dials in or meets in
person, so narrowing detection for everyone would trade one wrong behaviour for
another. Whether a link-less calendar entry is a meeting depends on how the user
works, which is why this is a setting and not a new default.

### Relationship to #375

#375 fixes the auto-stop half of the same problem: it stops a placeholder event
holding an already-running recording open, by requiring attributed meeting audio
or a room URL before matching on a calendar event. That guard is unconditional
because it is a correctness fix.

This PR is the detection half — whether a link-less entry should be treated as a
meeting at all — and is a genuine preference, so it is opt-in. The two are
independent and can land in either order.

## Validation

- `swift test --package-path native/MuesliNative` — 1546 tests in 150 suites,
  passing.
- Three new tests in `GoogleCalendarTests` (already in the meetings CI shard):
  link-less events are still selected by default; a placeholder is skipped when a
  link is required; and a linked event is still selected when a placeholder
  starts earlier and would otherwise win the sort.

Not covered by tests: the Settings toggle itself, which follows the existing
`settingsRow`/`settingsSwitch` pattern used by the neighbouring detection
controls and has no test precedent in this file.

## Contribution certification

      under Muesli's
      and licenses or terms.
      resulting changes.

### Third-party materials

None.

### AI assistance

Claude Code (Anthropic) was used materially: it traced the calendar selection
path, wrote the change and its tests, and ran the test suite. The change and the
test results were reviewed before submission.

---
<a href="https://app.blacksmith.sh/Muesli-HQ/codesmith/muesli/pr/394"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788953456&installation_model_id=422865&pr_number=394&repository=Muesli-HQ%2Fmuesli&return_to=https%3A%2F%2Fgithub.com%2FMuesli-HQ%2Fmuesli%2Fpull%2F394&signature=80a0c079a43bdcfcc9e6aa58bfbc486fa0f0466286ad93a7a70397772e5f2195"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a Meeting Notifications setting to require calendar events to include a meeting join link.
  * When enabled, events without join links are excluded from meeting notifications.
  * Events with valid join links are prioritized when selecting nearby calendar events.
  * The setting is saved and restored across app launches.

* **Tests**
  * Added coverage for meeting-link filtering, event selection, and setting persistence.

### DIFF
--- a/native/MuesliNative/Sources/MuesliNativeApp/Models.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/Models.swift
@@ -999,6 +999,11 @@ struct AppConfig: Codable {
     var showScheduledMeetingNotifications: Bool = true
     var scheduledMeetingNotificationLeadTime: ScheduledMeetingNotificationLeadTime = .atStart
     var showMeetingDetectionNotification: Bool = true
+    /// Only treat a calendar entry as a meeting when it carries a join link.
+    /// Off by default: link-less entries are still real meetings for people who
+    /// dial in or meet in person. On, it stops reminders and placeholders from
+    /// being detected as meetings at all.
+    var requireCalendarMeetingLink: Bool = false
     var mutedMeetingDetectionAppBundleIDs: [String] = []
     var meetingRecordingSavePolicy: MeetingRecordingSavePolicy = .never
     var meetingRecordingFileFormat: String = MeetingRecordingFileFormat.m4a.rawValue
@@ -1119,6 +1124,7 @@ struct AppConfig: Codable {
         case showScheduledMeetingNotifications = "show_scheduled_meeting_notifications"
         case scheduledMeetingNotificationLeadTime = "scheduled_meeting_notification_lead_time"
         case showMeetingDetectionNotification = "show_meeting_detection_notification"
+        case requireCalendarMeetingLink = "require_calendar_meeting_link"
         case mutedMeetingDetectionAppBundleIDs = "muted_meeting_detection_app_bundle_ids"
         case meetingRecordingSavePolicy = "meeting_recording_save_policy"
         case meetingRecordingFileFormat = "meeting_recording_file_format"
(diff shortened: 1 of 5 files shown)


jamesagarside:
Fair catch — the tests here covered the selection function but not the persisted
contract. Added in `c4dfc06b`:

- `require_calendar_meeting_link` added to the `AppConfig` encode/decode
  round-trip and to the snake_case key assertions.
- Missing key decodes to `false`. This is the one that matters: every config
  already on disk predates the key, and defaulting it on would silently stop
  detecting dial-in and in-person meetings.
- A wrong-typed value falls back to the default rather than failing the whole
  decode. `config.json` is hand-editable, so that behaviour is part of the
  contract — the decoder uses `try?` per key for exactly this reason.

Mutation-verified: flipping the default to `true` fails exactly the two decode
tests, and renaming the coding key fails exactly the snake_case test. Worth
noting the round-trip test survives the key rename by construction, since encode
and decode agree with each other — which is why the on-disk key needs its own
assertion rather than being assumed covered.

`ModelsTests` is not on `scripts/ci_unsharded_test_suites.txt`, so these run in
required CI.

Issue #322, opened 2026-07-15
Google Meet Microphone Inconsistency


cordemac:
When Muesli is enabled and recording prior to joining a Google Meet and I enter the Meet, the particpants on the other hand cannot hear me (no input to Google Meet through my microphone). The only way to resolve it is to exit the Google Meet, completely quit Muesli, then join the meet without the app running. Not sure if it's releated, but when I've joined Google Meet first, then open Muesli I sometimes get an issue where only records from the Google Meet speaker, not from my microphone. I'm on a Macbook Pro.

pHequals7:
Thanks for flagging this Corey - have experienced this myself as well - will try to fix this ASAP...

PR #396, opened 2026-08-11
Recover meeting mic on confirmed health degradation; episode-level telemetry

### LINKED ISSUE #381: Intermittent meeting mic route loss/mixing when Phone or WhatsApp call starts (macOS 26.6, M5)
## Summary

Muesli records the built-in microphone correctly when a meeting recording runs without a call. When a call starts in the macOS Phone app or WhatsApp, local microphone capture intermittently becomes silent, very quiet, or is replaced/mixed with system audio.

The remote participant still hears me clearly through the same MacBook microphone, so the microphone hardware and the call application's microphone access continue to work.

Muesli sometimes displays **“Your microphone is silent.”** In other cases, remote/system audio is incorrectly transcribed as `You`, while my actual local speech is missing or captured only intermittently.

An external earbud microphone currently provides a usable workaround.

## Environment

- Muesli: 0.8.1 (current release)
- Installation: official Homebrew cask
- macOS: 26.6 (25G72)
- Hardware: MacBook Air, Apple M5
- Meeting microphone tested with:
  - Automatic
  - MacBook Air Microphone
  - Huawei FreeBuds
- Stored built-in device UID when explicitly selected: `BuiltInMicrophoneDevice`
- Audio MIDI configuration during final tests: 48 kHz, approximately +5 dB input gain
- System audio capture configuration: CoreAudio tap enabled
- Call applications:
  - macOS Phone app
  - WhatsApp for macOS

## Expected behavior

Muesli should continue capturing:

- the local microphone as `You`
- call/system audio as the remote participant

Starting, stopping, or changing a call audio route should not make the local microphone disappear or cause system audio to be classified as `You`.

### A. Meeting recording without a call

1. Start a manual Muesli meeting recording.
2. Speak into the MacBook microphone.

Result:

- Local speech is recorded and transcribed as `You`.
- Input is relatively quiet but functional.

### B. Start Muesli first, then start a call

1. Start a Muesli meeting recording.
2. Confirm that local speech is transcribed correctly.
3. Start a call in Phone or WhatsApp.
4. Speak while the remote party or an automated hotline is speaking.
5. End the call and continue speaking.

Result:

- Before the call, local speech is captured correctly.
- During the call, local speech becomes very quiet, intermittent, or disappears.
- Remote/system audio is sometimes duplicated under both a remote speaker label and `You`.
- Local and remote sentences may be merged.
- After ending the call, local microphone capture often returns.

### C. Start the call first, then start Muesli

1. Start a call.
2. Confirm that the remote party can hear the MacBook microphone.
3. Start a Muesli meeting recording.

Result:

- Muesli may display “Your microphone is silent.”
- Only remote/system audio is recorded.
- The local voice can be completely absent from the saved audio and transcript.
- The remote party continues to hear the local voice normally.

### D. Switch input devices

1. Select Huawei FreeBuds as the meeting microphone.
2. Use the FreeBuds simultaneously in WhatsApp, or keep the MacBook microphone selected in WhatsApp.
3. Switch Muesli back to the built-in microphone during later tests.

Result:

- FreeBuds produce the most reliable and strongest local microphone capture.
- Speaker separation works substantially better with the FreeBuds.
- After switching to the FreeBuds and back, the built-in microphone sometimes begins working again, although at a lower level.
- The result is intermittent and difficult to reproduce consistently.
- Switching devices appears to reset or rebuild the input route temporarily.

## Diagnostic observations

- Muesli Dictation opens and records the built-in microphone successfully.
- Local diagnostics show that Dictation receives its first audio buffer and detects speech.
- Observed Dictation input levels were approximately -40 to -37 dB.
- The failure is specific to Meeting Recording while a call application is active.
- Affected saved recordings were checked in VLC and CapCut. The local voice is absent from the audio itself, so this is not only a transcription or diarization issue.
- Saved meeting records contain only the combined recording path; the microphone and system-audio path fields are empty.
- Switching between Automatic and the explicitly selected built-in microphone does not consistently resolve the issue.
- Changing the sample rate to 48 kHz and adjusting input gain does not resolve it.
- The transcription model performs well on the remote speaker, including German with a regional accent.
- The behavior depends on route state and device-switch history rather than one permanently unavailable microphone.

## Suspected route-change issue

Starting Phone or WhatsApp appears to switch macOS to a call/voice-processing audio route. The call application follows this route and continues receiving the microphone, but Muesli may:

- remain attached to the previous input route,
- receive a silent or low-level input stream,
- or receive system audio on the stream classified as `You`.

The fact that switching to an external microphone and back can temporarily restore the built-in microphone suggests an input-route lifecycle/recovery issue.

## Workaround

Selecting Huawei FreeBuds as the meeting microphone currently provides the most reliable result. Using separate physical microphones for the call application and Muesli is another option to test.

## Privacy

The original test calls included private account information, so I cannot attach them publicly. I can provide a new sanitized synthetic recording, redacted diagnostic output, or additional reproduction tests if needed.

pHequals7:
## Why

Production telemetry (0.8.1, since Aug 1) shows `meeting_microphone_capture_failed` as the dominant diagnostic error: 1,568 events across 92 users, ~17.6% affected-user prevalence in the latest 24h window, strongly power-law concentrated (top 5 users = 68% of events) with a macOS 26 skew. Analysis in `Context/handoff-2026-08-11-repeated-mic-failure-errors.md`.

Two compounding defects:

1. **Detection never connected to recovery.** `MeetingMicHealthTracker` confirms a semantically dead mic graph (≥3s of missing or all-zero mic callbacks while system audio is active), but `RouteAwareMeetingMicRecorder` only recovered when a child recorder *threw*. A graph that silently stops delivering real samples never throws — so the user's side of the transcript silently dies. Issue #381 is the canonical live report (call starts → mic degrades while the call app keeps the mic).
2. **Telemetry inflates and anonymizes it.** Every warning flap emitted a full error event, collapsing three distinct internal reasons into one generic error.

## What changes

- **New `MeetingMicRecoveryCoordinator`**: turns the raw health-state stream into one episode per degradation (`degraded` → `recovered` / `unrecovered`), counting flaps internally, and drives **bounded same-route recovery**: one immediate attempt at episode start (the tracker already confirmed for ~3s), cooldown-limited retries (15s), capped at 3 attempts per episode. Recovery uses the existing safe handoff primitive — the active recorder keeps capturing while a candidate is prepared, and the candidate promotes only after producing a non-empty buffer.
- **`RouteAwareMeetingMicRecorder.requestSameRouteRecovery(reason:)`**: forces the existing serialized handoff for the *current* route. No-op (returns false) when not recording or a handoff is already pending. Protocol default no-op for other recorders.
- **`MeetingSession`**: feeds every health snapshot to the coordinator; closes episodes at meeting stop.
- **`MuesliController` telemetry**: per-flap error emission removed. Now: `meeting.microphone.degraded` and `meeting.microphone.recovered` signals once per episode (with reason, state, duration, flap count, recovery attempts), and `meeting_microphone_capture_failed` is reserved for meetings that **end unrecovered** — the terminal condition. User-facing warning UI unchanged.

Privacy: episode fields are state/reason enums and counts only — no audio, device names, titles, or content.

## Validation

- 10 new coordinator tests (episode lifecycle, flaps, cooldown/cap, recovery-once semantics, terminal-on-stop, quiet-room negative) + 3 new recorder tests (health-triggered same-route promotion, no-op when not recording, no stacked pendings). Drives the real `MeetingMicHealthTracker` with synthetic sample streams.
- Full suite: **1567/1567 green**.
- Local live validation (Phone/WhatsApp call-transition protocol from the Aug 6 investigation) is staged but pending dev-lane availability.

## Contribution Certification

## Third-Party Materials

None.

## AI Assistance

Developed with AI assistance (OpenAI Codex): design, implementation, and tests. The production analysis motivating this change is recorded in the referenced Context handoff.

---
<a href="https://app.blacksmith.sh/Muesli-HQ/codesmith/muesli/pr/396"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789021674&installation_model_id=422865&pr_number=396&repository=Muesli-HQ%2Fmuesli&return_to=https%3A%2F%2Fgithub.com%2FMuesli-HQ%2Fmuesli%2Fpull%2F396&signature=34d953b52ec12eefa4a9d35b5d4b4087b0d9b4cc92da6ac1eb63082f5f605ade"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added automatic microphone recovery when audio health degrades.
  * Added episode-level reporting for degraded, recovered, and unrecovered states.
  * Recovery attempts now use cooldowns and limits to prevent repeated interruptions.
  * Health episodes include recovery status, duration, and contributing details.
  * Added visibility into microphone handoff outcomes, including successful, failed, and timed-out recoveries.

* **Bug Fixes**
  * Improved same-route recorder handoffs while preserving active audio until replacement is available.
  * Reduced duplicate recovery attempts and unnecessary diagnostic warnings.
  * Meeting completion now closes active microphone health episodes cleanly.

### DIFF
--- a/native/MuesliNative/Sources/MuesliNativeApp/AudioQueueInputRecorder.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/AudioQueueInputRecorder.swift
@@ -86,6 +86,17 @@ final class AudioQueueInputRecorder: StreamingDictationRecording, StreamingDicta
         emitLatency("audio_queue_start_begin")
         let status = AudioQueueStart(audioQueue, nil)
         emitLatency("audio_queue_start_end")
+        // AudioQueueStart can block while the daemon negotiates the route, and
+        // teardown may land during that window. If this instance was invalidated
+        // while the call was in flight, synchronously stop what just started
+        // before returning — capture must not outlive teardown.
+        if permanentlyInvalidated {
+            AudioQueueStop(audioQueue, true)
+            isRunning = false
+            captureGeneration &+= 1
+            cleanupAfterStartFailure()
+            throw Self.runtimeError(code: 9, message: "Recorder was invalidated by teardown")
+        }
         guard status == noErr else {
             isRunning = false
             captureGeneration &+= 1
--- a/native/MuesliNative/Sources/MuesliNativeApp/AudioQueueInputRecorder.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/AudioQueueInputRecorder.swift
@@ -15,7 +15,10 @@ final class AudioQueueInputRecorder: StreamingDictationRecording, StreamingDicta
 
     private let directoryName: String
     private let queueLock = NSRecursiveLock()
-    private var permanentlyInvalidated = false
+    /// Published independently of queueLock: invalidateForTeardown() must land
+    /// even while a worker is blocked inside AudioQueueStart holding queueLock,
+    /// so the post-start self-check observes it before start() returns.
+    private let teardownInvalidation = OSAllocatedUnfairLock(initialState: false)
     private let stateLock = OSAllocatedUnfairLock(initialState: FileState())
     private let processingQueue = DispatchQueue(label: "com.muesli.audio-queue-input-recorder-processing")
     private let failureCallbackQueue = DispatchQueue(label: "com.muesli.audio-queue-input-recorder-failures")
@@ -47,7 +50,7 @@ final class AudioQueueInputRecorder: StreamingDictationRecording, StreamingDicta
     func prepare() throws {
         queueLock.lock()
         defer { queueLock.unlock() }
-        guard !permanentlyInvalidated else {
+        guard !isPermanentlyInvalidated else {
             throw Self.runtimeError(code: 9, message: "Recorder was invalidated by teardown")
         }
 
@@ -57,7 +60,7 @@ final class AudioQueueInputRecorder: StreamingDictationRecording, StreamingDicta
     func start() throws {
         queueLock.lock()
         defer { queueLock.unlock() }
-        guard !permanentlyInvalidated else {
+        guard !isPermanentlyInvalidated else {
             throw Self.runtimeError(code: 9, message: "Recorder was invalidated by teardown")
         }
 
@@ -90,7 +93,7 @@ final class AudioQueueInputRecorder: StreamingDictationRecording, StreamingDicta
         // teardown may land during that window. If this instance was invalidated
         // while the call was in flight, synchronously stop what just started
         // before returning — capture must not outlive teardown.
-        if permanentlyInvalidated {
+        if isPermanentlyInvalidated {
             AudioQueueStop(audioQueue, true)
             isRunning = false
             captureGeneration &+= 1
@@ -186,9 +189,11 @@ final class AudioQueueInputRecorder: StreamingDictationRecording, StreamingDicta
     /// never start capture again, regardless of which queue a stale worker is
     /// on. Distinct from cancel(), which disposes but permits re-prepare.
     func invalidateForTeardown() {
-        queueLock.lock()
-        permanentlyInvalidated = true
-        queueLock.unlock()
+        teardownInvalidation.withLock { $0 = true }
+    }
+
+    private var isPermanentlyInvalidated: Bool {
+        teardownInvalidation.withLock { $0 }
     }
 
     func pause() {
--- a/native/MuesliNative/Sources/MuesliNativeApp/AudioQueueInputRecorder.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/AudioQueueInputRecorder.swift
@@ -15,6 +15,7 @@ final class AudioQueueInputRecorder: StreamingDictationRecording, StreamingDicta
 
     private let directoryName: String
     private let queueLock = NSRecursiveLock()
+    private var permanentlyInvalidated = false
     private let stateLock = OSAllocatedUnfairLock(initialState: FileState())
     private let processingQueue = DispatchQueue(label: "com.muesli.audio-queue-input-recorder-processing")
     private let failureCallbackQueue = DispatchQueue(label: "com.muesli.audio-queue-input-recorder-failures")
@@ -46,6 +47,9 @@ final class AudioQueueInputRecorder: StreamingDictationRecording, StreamingDicta
     func prepare() throws {
         queueLock.lock()
         defer { queueLock.unlock() }
+        guard !permanentlyInvalidated else {
+            throw Self.runtimeError(code: 9, message: "Recorder was invalidated by teardown")
+        }
 
         try prepareLocked()
     }
@@ -53,6 +57,9 @@ final class AudioQueueInputRecorder: StreamingDictationRecording, StreamingDicta
     func start() throws {
         queueLock.lock()
         defer { queueLock.unlock() }
+        guard !permanentlyInvalidated else {
+            throw Self.runtimeError(code: 9, message: "Recorder was invalidated by teardown")
+        }
 
         guard !isRunning else { return }
         try prepareLocked()
@@ -162,6 +169,15 @@ final class AudioQueueInputRecorder: StreamingDictationRecording, StreamingDicta
 
     func currentPower() -> Float {
         stateLock.withLock { $0.latestPowerDB }
+    }
+
+    /// Terminal and synchronous: after meeting teardown, this instance must
+    /// never start capture again, regardless of which queue a stale worker is
+    /// on. Distinct from cancel(), which disposes but permits re-prepare.
+    func invalidateForTeardown() {
+        queueLock.lock()
+        permanentlyInvalidated = true
+        queueLock.unlock()
     }
 
     func pause() {
--- a/native/MuesliNative/Sources/MuesliNativeApp/FallbackStreamingDictationRecorder.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/FallbackStreamingDictationRecorder.swift
@@ -120,6 +120,13 @@ final class FallbackStreamingDictationRecorder: StreamingDictationRecording, Str
         wireCallbacks()
     }
 
+    func invalidateForTeardown() {
+        lock.lock()
+        defer { lock.unlock() }
+        primary.invalidateForTeardown()
+        fallback.invalidateForTeardown()
+    }
+
     func pause() {
         lock.lock()
         let recorder = activeRecorderLocked()
--- a/native/MuesliNative/Sources/MuesliNativeApp/MeetingMicRecording.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/MeetingMicRecording.swift
@@ -26,6 +26,16 @@ struct MeetingMicRecorderDiagnosticsSnapshot: Codable, Equatable {
     let route: MeetingMicRouteDiagnosticsSnapshot?
 }
 
+/// Result of a same-route recovery request. The distinction matters to the
+/// coordinator's accounting: a busy recorder is back-pressure (in-flight
+/// handoff already covers the episode), while an unavailable recorder cannot
+/// recover at all and the request must count toward the episode's attempt cap.
+enum MeetingMicRecoveryRequestResult: Equatable {
+    case initiated
+    case busy
+    case unavailable
+}
+
 protocol MeetingMicRecording: AnyObject {
     var preferredInputDeviceID: AudioObjectID? { get set }
     var onRawPCMSamples: (([Int16]) -> Void)? { get set }
@@ -52,16 +62,15 @@ protocol MeetingMicRecording: AnyObject {
     /// graph for the current input while the active graph keeps running, and
     /// promote it only after it produces a non-empty buffer. Used when the
     /// health tracker confirms a semantically dead graph (missing/zero
-    /// callbacks) that never raises an explicit CoreAudio error. Returns true
-    /// when a recovery handoff was actually initiated. Recorders without a
-    /// recovery primitive keep the default no-op (returns false).
+    /// callbacks) that never raises an explicit CoreAudio error. Recorders
+    /// without a recovery primitive keep the default no-op (.unavailable).
     @discardableResult
-    func requestSameRouteRecovery(reason: String) -> Bool
+    func requestSameRouteRecovery(reason: String) -> MeetingMicRecoveryRequestResult
 }
 
 extension MeetingMicRecording {
     func invalidateForTeardown() {}
-    func requestSameRouteRecovery(reason: String) -> Bool { false }
+    func requestSameRouteRecovery(reason: String) -> MeetingMicRecoveryRequestResult { .unavailable }
 }
 
 final class StreamingMeetingMicRecorderAdapter: MeetingMicRecording {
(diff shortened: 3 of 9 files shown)


pHequals7:
Both shutdown-race findings are addressed in 2f18b5bf + 65a7c78a:

**Greptile P1 / CodeRabbit (dangling episode on shutdown):** fixed twice over. (1) `finishMeeting()` moved after the `chunkRotationQueue` barrier in `stop()`, so every sample callback enqueued before teardown is processed before finalization and later callbacks bail on `isRecording == false`. (2) The coordinator now seals itself: after `finishMeeting()`, `process()\) is ignored, so a late callback cannot open a dangling episode regardless of caller ordering. Regression test: `lateSnapshotsAfterFinishMeetingAreIgnored` (degraded → finish → further degraded snapshots produce nothing).

**CodeRabbit 🟠 (callbacks under the coordinator lock):** fixed. Attempt reservations and episode transitions commit under the lock; `recoveryRequest` and `onEpisodeEvent` now fire after unlock, so synchronous re-entry cannot deadlock and the real-time sample queue is never stalled by a callback. Uninitiated reservations roll back. Regression tests: `episodeCallbackReentryDoesNotDeadlock`, `recoveryCallbackReentryDoesNotDeadlock`.

Full suite: 1570/1570 green locally on macOS 26.5.2 (the T-Rex Linux host lacks a Swift toolchain — happy to attach local run output if useful).

pHequals7:
All four blocking findings addressed in `30e06418` (rebased onto current main):

**1. Terminal failures silently dropped (P1)** — confirmed and fixed: `stopMeetingRecording()` clears `activeMeetingID` before `MeetingSession.stop()` runs, so the episode gate rejected the terminal event. The controller now tracks `lastStoppedMeetingID` and the gate accepts the most-recently-stopped session. The terminal error can no longer be swallowed by the stop ordering, so the post-release metric cannot 'fall to zero because of a guard.'

**2. One healthy buffer reset the retry budget (P1)** — fixed: episodes now close as *recovered* only after **10s of sustained healthy delivery**. A degraded→blip→degraded flap stays in the same episode with its original attempt budget (new test: `healthyBlipKeepsEpisodeOpen`). Meeting end after a healthy-but-unsustained blip closes as recovered; still-degraded closes as unrecovered (`meetingEndAfterHealthyBlipIsRecovered`).

**3. Terminal telemetry discarded its classification (P1)** — fixed: episode events now carry `episode_id`, `recorder_kind`, `route_category`, `selected_input_resolved`, `handoff_promotions`, `last_handoff_outcome` (promoted/failed/timed_out via a new recorder outcome callback), and `recovery_credited` (a promoted handoff preceded the healthy transition). The unrecovered path emits the full parameter set as `meeting.microphone.unrecovered` *and* keeps the legacy `meeting_microphone_capture_failed` incident for dashboard continuity. So post-release we can distinguish missing-callbacks vs zero-buffer, see whether the restart repaired it, and compare cohorts — including the macOS 26 slice.

**4. Reservation identity (P2)** — reservations now carry a unique UUID token stored on the episode, revalidated immediately before dispatch (a closed/replaced episode invalidates it), with token-checked settlement/rollback. Regression test: `closedEpisodeInvalidatesPendingRecovery`.

Full suite 1601/1601 green locally (macOS 26.5.2); focused suites re-verified after the rebase.

pHequals7:
P1 from the latest Greptile review (discard path) — fixed in 5cc614f2: `discardMeetingRecording` cleared `activeMeetingID` immediately after `session.discard()`, while the terminal episode event hops to the main actor asynchronously, so discarded-unrecovered meetings lost both the terminal signal and the incident. Both discard paths now set `lastStoppedMeetingID` before clearing the active identity, matching the stop() fix. (CodeRabbit's quiet-room test note is also in: the negative test now feeds >3s of zero mic samples, past the detector's confirmation window.)

pHequals7:
Security P1 (queued recovery could restart capture after teardown) — fixed in 3c49b845, and thank you, that was a real one: the recovery handoff is the first path that can be *queued near meeting end*, so the old assumption (handoffs only happen mid-meeting) stopped holding.

The worker now revalidates candidate ID + generation + running lifecycle **immediately before `prepare()` and again before `start()`**; `stop()`/`cancel()` clear the pending candidate and bump the generation, so a queued worker that executes after teardown returns without touching audio — and a candidate whose prepare was interrupted mid-flight is cancelled rather than left half-built. Regression test: `stopWithQueuedRecoveryNeverStartsCandidate` blocks the worker inside `prepare()`, stops the meeting, releases the gate, and asserts the candidate never starts.

Full suite: 1603/1603 green locally (macOS 26.5.2).

pHequals7:
Second-round security P1 (TOCTOU between guard and start) — fixed in 464aea4a by inverting the race rather than tightening it:

The worker-side revalidation could never be fully airtight because `start()` can block and teardown's cancellation was async. So teardown now **poisons synchronously**: `stop()`/`cancel()` call `invalidateForTeardown()` on every child (active/pending/unused) before any async cleanup, and an invalidated recorder instance permanently refuses `prepare()`/`start()` (distinct from `cancel()`, which stays reusable for the dictation lifecycle). Whichever way the guard/start/teardown interleave, invalidation wins — there is no window in which a stale worker can start capture after meeting end.

The regression test now asserts the pending candidate is poisoned *synchronously during stop()* — before its blocked prepare is even released — and never starts. Full suite 1603/1603 green.

pHequals7:
Both P1s from the 2/5 round addressed in 46b879b9:

**1. Capture entering startup when teardown lands.** Both recorders now re-check the teardown-invalidation flag *the moment their blocking start call returns*, and synchronously stop + discard what just started (`AudioQueueStop`/`stopEngineSafely` + temp-file removal) before `start()\) returns. Combined with the synchronous poisoning from the previous commit, the residual exposure is bounded to inside the daemon's own start call — the client-side minimum.

**2. Single-slot `lastStoppedMeetingID` overwrite.** Replaced by `RecentMeetingIdentityGate`: a bounded (8, FIFO-evicted) authorized-ID set. Sessions are authorized at creation and again at stop/discard; a queued terminal callback from an earlier stopped meeting is no longer dropped when a later meeting stops first. Covered by new unit tests (eviction, multi-stopped authorization, idempotent re-auth).

Full suite: 1606/1606 green locally.

pHequals7:
Third-round security P1 — fixed in c4772adb. The lock analysis was the real bug: both recorders published `permanentlyInvalidated` behind the same lock their blocking `start()` holds, so teardown's invalidation queued behind a stuck engine start and the post-start self-check read a stale flag.

The invalidation flag now lives in its own `OSAllocatedUnfairLock`, independent of `queueLock`/`graphLock`. Teardown publishes it immediately even while a worker is blocked inside `AudioQueueStart`/engine startup, and the post-start self-check — which runs before `start()` returns — observes it and synchronously stops + discards the just-started capture. Net guarantee: a candidate either never starts (entry guard), or self-stops before its start returns (post-check), or was legitimately running before teardown began and is cancelled/stopped by the normal teardown path.

Suite: 1606/1606 green locally.

pHequals7:
Round-4 item fixed in 8e9c6e49: teardown now calls `stop()` on a pending handoff candidate **synchronously** — capture cessation is bounded and fast — and deletes its scratch file, while disposal (`cancel()`, the potentially blocking part) stays on the async cleanup queue. So the guarantee is now complete across all interleavings:

- not yet started at teardown → poisoned entry guards reject `prepare()`/`start()`;
- blocked inside start when teardown lands → invalidation publishes independently of the startup lock; the post-start self-check self-stops before returning;
- fully started before teardown → synchronous `stop()\) in the teardown path itself;
- disposal of resources → async (deliberately — that path can block on wedged CoreAudio cleanup and must not stall meeting teardown).

(The local full-suite run had one failure in `Diarizer preload coordination` — a timing-sensitive suite unrelated to this diff; it passes standalone. Flagging it as a flake to watch rather than mixing it into this PR's scope.)

pHequals7:
P1 from the latest Greptile round (recovery-refusal resets the retry budget) — fixed in d1021849, and their harness evidence was decisive: a transient refusal (the normal 'handoff already pending' back-pressure) rolled back `lastAttemptAt`, so sustained degradation re-dispatched a recovery request per snapshot (their run: 100 snapshots → 100 requests).

Now a declined request refunds the attempt budget but **keeps** the back-pressure timestamp, so the cooldown still throttles re-dispatch. Regression test `transientRefusalsRespectCooldown` feeds ~19s of continuous degradation through a refusing recorder and asserts exactly one request per cooldown window (2 total), not one per snapshot.

Note: branch was rebased onto the PR's current remote head (which includes main via 95f6d4da) before pushing. Full suite: 1607/1607 green locally.

pHequals7:
Added one scoped addition from live validation: **user-muted input classification** (e6b31017).

During lane validation we confirmed the detector cannot distinguish a muted/zero-gain input from a broken route — both are all-zero samples with active system audio, and the founding incident (June 7 all-zero meeting) is exactly why the detector exists. Without classification, a user-muted mic would open recovery episodes and eventually a terminal error, polluting the new telemetry with user intent.

Now: at episode confirmation (once — never per sample), the capture device's input gain/mute state is read via HAL; muted input suppresses the episode, emits `meeting.microphone.user_muted` once per meeting, and spends zero recovery work. Un-muting while still degraded opens the episode normally, so a genuinely broken route is never masked. The 'microphone is silent' banner stays — it is accurate for a muted user.

Tests: muted-suppression, suppression-while-degraded, unmute-opens-episode, no-latch-after-healthy. Full suite 1623/1623 green.

pHequals7:
Round-5 items fixed in 40121c36:

**1. Synchronous pending stop can stall teardown (P1)** — correct catch, and it conflicts with round 4's fix; reconciled per your recommendation: the pending candidate's synchronous `stop()` is reverted to async disposal. Capture cessation is guaranteed without the lock contention via the two earlier layers: entry guards reject start after invalidation, and the post-start self-check (which reads the independently-published invalidation flag) self-stops any start that was in flight through teardown — before it returns. Teardown never waits on the startup lock again.

**2. Refusals permanently evade the attempt cap (P1)** — fixed with a three-way result instead of a Bool: `requestSameRouteRecovery` now returns `.initiated` / `.busy` (handoff already pending — refunded attempt, cooldown timestamp retained for back-pressure) / `.unavailable` (counted toward the cap). So transient back-pressure can't churn per-snapshot, AND a perpetually-unavailable recorder hits the 3-attempt cap instead of dispatching once per cooldown forever. Updated tests: `unavailableRecoveryCountsTowardCap` (cap binds refusals), `transientRefusalsRespectCooldown` (.busy → one dispatch per cooldown window).

Local full suite: 1623/1623 (the one failure in that run was `PasteController` clipboard-simulation — unrelated to this diff, passes standalone; flagging as another flake to watch).

pHequals7:
The 1/5 review round on 40121c36 appears to have analyzed a stale tree — four of its five findings describe code that commit replaced. Verified against the current head:

1. **"Recovery request that returns `false` clears its reservation"** — the Bool API no longer exists. `requestSameRouteRecovery` returns `.initiated/.busy/.unavailable`; `.unavailable` retains the attempt count (cap binds refusals), `.busy` refunds but keeps the cooldown timestamp. See `MeetingMicRecoveryCoordinator.dispatchRecovery` and tests `unavailableRecoveryCountsTowardCap` / `transientRefusalsRespectCooldown`.
2. **"stop()/cancel() synchronously call pending.recorder.stop()"** — removed in this very commit; pending disposal is async via `cancelAsync` (`MeetingMicRecording.swift:337-348`), precisely to avoid the graphLock contention this round also flags.
3. **"invalidateForTeardown() only sets a flag under graphLock/queueLock"** — the flag now lives in a dedicated `OSAllocatedUnfairLock` in both recorders (`AudioQueueInputRecorder.swift:21`, `StreamingMicRecorder.swift:73`), so teardown invalidation publishes even while a worker is blocked in a synchronous start.
4. **"post-start cleanup only after startup returns"** — that is the designed self-stop: a start blocked through teardown observes the independently-published flag at return and synchronously stops+discards capture before `start()\) returns.

The one residual this leaves is deliberate and worth stating explicitly: a candidate that *completed* startup before teardown began was legitimately capturing while the meeting was live; its disposal rides the async cleanup queue (milliseconds), because synchronous cancel can block on wedged CoreAudio cleanup (#322-era, ~30s observed) and meeting teardown must never hang behind it. If the team prefers a hard synchronous bound there, the alternative is a watchdog-forced dispose — happy to add it, but it trades a ms-scale cleanup lag for a potential multi-second teardown hang.

Full suite: 1623/1623 green locally. @greptile please re-review against the current head — the T-Rex artifacts on this round reference superseded implementations.

pHequals7:
Superseded by #406 — same final code squashed onto current main, with the full hardening rationale folded into a fresh description. Closing this one to retire the review archaeology. All findings from the rounds here are addressed in the commits squashed into that PR (Greptile reached 5/5 on this branch's final iteration before the supersede).

PR #406, opened 2026-08-13
Recover meeting mic on confirmed health degradation; episode-level telemetry

pHequals7:
## Why

Production telemetry on 0.8.1 (analysis: `Context/handoff-2026-08-11-repeated-mic-failure-errors.md`): `meeting_microphone_capture_failed` is the dominant diagnostic error — 1,568 events / 92 users in 10 days, ~17.6% affected-user prevalence in the latest 24h window, power-law concentrated (top 5 users = 68%), macOS 26-skewed. Issue #381 is the canonical report: a Phone/WhatsApp call starts, the mic graph keeps "running" but delivers silence, and the user loses their side of the transcript.

Two compounding defects:
1. **Detection never connected to recovery.** `MeetingMicHealthTracker` confirms a dead mic graph, but `RouteAwareMeetingMicRecorder` only recovered on explicit recorder errors — and a semantically-dead graph never throws.
2. **Telemetry inflated and anonymized it.** Every warning flap emitted an error; three distinct internal reasons collapsed into one generic event.

## What changes

- **`MeetingMicRecoveryCoordinator`** (new): collapses health-state flapping into one episode per degradation. Recovery: immediate first attempt after the tracker's ~3s confirmation, 15s cooldown, max 3 attempts per episode. Episodes close as *recovered* only after **10s of sustained healthy delivery** — a healthy blip on a flapping route never resets the retry budget. Meeting end while degraded = `unrecovered` (terminal); healthy-but-unsustained at end = `recovered`.
- **`requestSameRouteRecovery(reason:)`** on `RouteAwareMeetingMicRecorder`: reuses the existing serialized handoff — the active recorder keeps capturing while a candidate prepares, and the candidate promotes only after producing a non-empty buffer. Returns `.initiated / .busy / .unavailable` so coordinator accounting separates back-pressure (refunded attempt, cooldown retained) from genuine unavailability (counts toward the cap).
- **Teardown safety** (recovery handoffs can be queued near meeting end — a new exposure): children are poisoned synchronously via `invalidateForTeardown()`, published on a dedicated lock independent of the startup critical section; a start blocked through teardown self-stops and discards before returning; handoff workers revalidate candidate identity + generation + lifecycle before prepare and before start. Capture can never begin or outlive meeting teardown.
[...]

PR #408, opened 2026-08-13
Recover meeting system audio on tap failure; heartbeat watchdog + bounded rebuild retry

pHequals7:
## Why

Live field evidence from this week (analysis: `Context/handoff-2026-08-13-pr406-meeting-mic-recovery-merged.md`): during a real Google Meet recorded on a dev build, casing AirPods mid-meeting made the system-audio tap's route-change rebuild fail once (`tapCreationFailed(0)`) — and the tap never came back. **The remaining ~40% of the meeting lost the remote side entirely**: no retry, no error event, no recovery, and the mic health tracker went blind at the same moment (its degradation precondition is *active system audio*).

The restart path today gets exactly one attempt; on any error it nils `onPCMSamples`, marks itself not-recording, and logs to stderr. That is the whole failure story.

## What changes

1. **Bounded retry on route-rebuild failure.** `CoreAudioSystemRecorder` retries a failed rebuild on a tested backoff schedule (0.5s/1.5s/3.5s — sized to the observed route-churn window), superseding pending retries when newer route changes arrive, and only after exhaustion marks capture dead *and reports it* (`onCaptureFailure`). Retries are cancelled on `stop()`.
2. **Heartbeat-based tap health.** The IO callback advances a monotonic counter (it fires continuously regardless of content — a stall means a dead graph, quiet rooms included). Exposed as `captureHeartbeat` on `SystemAudioCapturing` (SCK fallback exempt by default).
3. **`MeetingSystemAudioWatchdog`** (new, driven by a 1s timer from `MeetingSession`): opens a tap-health episode on heartbeat stall (>2s) or terminal failure, drives cooldown-limited rebuilds (15s, max 4/episode) via `rebuildForHealthRecovery`, closes after 2 sustained alive ticks, terminalizes at stop/discard. Paused meetings and rebuild-in-flight windows don't tick.
4. **Mic-tracker blindness bridge.** While the tap is dead, mic-callback staleness (>3s) alone opens a mic episode via `MeetingMicRecoveryCoordinator.noteExternalDegradation` — today the mic detector goes blind exactly when its sibling pipeline dies (this is what masked the live cascade).
5. **Telemetry**: `meeting.system_audio.degraded / .recovered / .unrecovered` signals once per episode (reason, duration, attempts), and a new `meeting_system_audio_capture_failed` incident kind for meetings that end with the tap dead. The session-identity gate from #406 is reused for terminal callbacks.

## Validation

- 12 new watchdog tests (stall/recovery/cooldown-cap/terminalization/pause/blindness-bridge semantics with injected clock + heartbeat) + 3 coordinator external-degradation tests + retry-policy backoff test.
- **Full suite: 1639/1639 green** (macOS 26.5.2). (Two unrelated timing flakes — PasteController, Diarizer preload — recur under full-suite load, pass standalone, untouched by this diff.)
- The failure this PR addresses was observed live on a dev lane with daemon-log ground truth; the retry path will be re-validated with an AirPods case/out cycle on a lane build before or after merge (lane availability permitting).

## Contribution Certification

## Third-Party Materials

None.

## AI Assistance

Developed with AI assistance (OpenAI Codex): design, implementation, tests. The motivating failure was captured from a real meeting on macOS 26.5.2 with coreaudiod unified-log evidence.

---
<a href="https://app.blacksmith.sh/Muesli-HQ/codesmith/muesli/pr/408"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789242303&installation_model_id=422865&pr_number=408&repository=Muesli-HQ%2Fmuesli&return_to=https%3A%2F%2Fgithub.com%2FMuesli-HQ%2Fmuesli%2Fpull%2F408&signature=263536adefa12654b627a7e9bfda745e07ac4c61da50a75806f790c320d3a08e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added system-audio capture health monitoring during meetings.
  * Automatically recovers from capture stalls, route changes, and transient failures.
  * Reports degraded, recovered, and unrecovered capture states.
  * Connects system-audio failures with microphone recovery workflows.
* **Bug Fixes**
  * Prevented unnecessary recovery attempts during pauses, route transitions, or after meetings end.
  * Added bounded retries and clearer handling for unrecoverable failures.
  * Improved microphone stability with debounced configuration restarts and a longer route handoff window.
* **Diagnostics**
  * Added incident tracking for system-audio capture failures.

### DIFF
--- a/native/MuesliNative/Sources/MuesliNativeApp/CoreAudioSystemRecorder.swift
+++ b/native/MuesliNative/Sources/MuesliNativeApp/CoreAudioSystemRecorder.swift
@@ -32,6 +32,10 @@ protocol SystemAudioCapturing: AnyObject {
     /// Whether this backend advances captureHeartbeat during capture. Backends
     /// without a heartbeat (SCK) must be excluded from stall monitoring.
     var supportsHeartbeatMonitoring: Bool { get }
+    /// True while a route transition is still settling (recent notification).
+    /// The watchdog skips stall evaluation in this window: the old tap's
+    /// heartbeat stalling mid-transition is expected, not a dead graph.
+    var isRouteSettling: Bool { get }
     /// Health-driven rebuild: tear down and recreate the capture graph with
     /// the same bounded retry policy used for route changes. Returns whether a
     /// rebuild was actually started. Default no-op (false) for backends without
@@ -44,6 +48,7 @@ extension SystemAudioCapturing {
     var captureHeartbeat: UInt64 { 0 }
     var isRebuilding: Bool { false }
     var supportsHeartbeatMonitoring: Bool { false }
+    var isRouteSettling: Bool { false }
     var onCaptureFailure: ((Error) -> Void)? {
         get { nil }
         set {}
@@ -120,6 +125,18 @@ final class CoreAudioSystemRecorder: SystemAudioCapturing, SystemAudioDiagnostic
     /// Settle debounce for route-change rebuilds: how long after the last
     /// route notification before the rebuild fires.
     static var routeSettleDelay: TimeInterval = 1.5
+    /// Last default-output route notification (ms since epoch; 0 = never).
+    /// Read across queues via atomics; written from processingQueue.
+    private let lastRouteChangeAtMs = ManagedAtomic<Int64>(0)
+    /// True while a route transition is still settling (recent notification).
+    /// The watchdog skips stall evaluation in this window: the old tap's
+    /// heartbeat stalling mid-transition is expected, not a dead graph.
+    var isRouteSettling: Bool {
+        let lastRoute = lastRouteChangeAtMs.load(ordering: .relaxed)
+        guard lastRoute != 0 else { return false }
+        let elapsedMs = Date().timeIntervalSince1970 * 1000 - Double(lastRoute)
+        return elapsedMs < (Self.routeSettleDelay * 1000 + 2000)
+    }
     /// Test seam for the rebuild path (HAL create+start is not unit-testable).
     var createAndStartForTesting: (() throws -> Void)?
     private(set) var isRecording: Bool {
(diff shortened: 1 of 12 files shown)


pHequals7:
Both round-1 P1s fixed in 2d448f3d:

**1. SCK fallback always stalled (P1)** — the ScreenCaptureKit backend reports heartbeat 0 permanently, so the watchdog would have fired false degraded episodes + recovery requests on every SCK meeting. `SystemAudioCapturing` now declares `supportsHeartbeatMonitoring` (CoreAudio tap: true; SCK: false via default), and the watchdog timer only starts for heartbeat-capable backends. Test: `heartbeatCapabilityByBackend`.

**2. Terminal failure rejected watchdog recovery (P1)** — real design flaw your probe executed cleanly: exhaustion flipped `isRecording` before `onCaptureFailure`, so the watchdog's rebuild was rejected by the guard and the tap stayed dead all meeting. Exhaustion now keeps the recorder alive with a `captureDead` flag instead; `rebuildForHealthRecovery` stays accepted from that state and clears it on success. Executable coverage via a test seam driving the rebuild path with injected failures: `terminalFailureRemainsRecoverable` (exhaust → dead → watchdog rebuild → recovered) and `rebuildRetriesThenSucceeds` (backoff then success, no terminal).

Full suite: 1642/1642 green locally (macOS 26.5.2).

pHequals7:
Final architecture after a full day of live transition testing on macOS 26.5.2 (degraded 6-day daemon AND fresh daemon):

**The headline change: the route-change rebuild is gone.** Git archaeology showed it was added May 10 (8319c3ba) when the tap was bound to the output device's stream, and survived the May 21 switch to the route-independent global process mix (b639951f) as a vestige. Live proof on this PR's build: an AirPods connect/case cycle with the listener disabled produced zero rebuild events and zero content loss — so the listener is now record-only (it timestamps transitions for the watchdog's settle window) and never rebuilds.

**Mic side settles too**: the recovery coordinator defers dispatches while a transition is settling (`isRouteSettling`), and `StreamingMicRecorder`'s config-change restart is debounced (1.5s settle) instead of firing mid-notification-burst; handoff candidate timeout relaxed 2s→5s for slow BT bring-up on aged daemons.

**Measured outcome**, same host, same test, same day: degraded daemon → transition self-heals in ~1 min with zero manual intervention (previously: permanent tap death or a minutes-long rebuild storm with daemon overloads + a frozen unrelated YouTube). Fresh daemon → the transition is a non-event: one log line, zero failures, zero overloads.

Full suite: 1646/1646 relevant green locally (PasteController/StreamingVadController flakes pass standalone; untouched by this diff).

pHequals7:
The two P1s in the latest round describe code that no longer exists on this head — they were fixed in 2d448f3d, which is an ancestor of the reviewed commit 4c2ad181. Verified against the current tree:

1. **"SCK fallback always stalled"** — `MeetingSession.swift:1132` gates the watchdog timer on `supportsHeartbeatMonitoring` (protocol requirement; `false` default for SCK at `CoreAudioSystemRecorder.swift:50`, `true` for the tap at line 112). An SCK meeting never starts the watchdog. Covered by `heartbeatCapabilityByBackend`.
2. **"Terminal failure clears isRecording before onCaptureFailure"** — the exhaustion path no longer touches `isRecording`/`onPCMSamples`; it sets `captureDeadFlag` (`CoreAudioSystemRecorder.swift:942`) and stays recoverable, and `terminalFailureRemainsRecoverable` drives exactly exhaust → watchdog rebuild → success through a test seam.

The T-Rex artifacts on this round quote the pre-2d448f3d implementations. @greptile please re-review against 4c2ad181 — and note this is the second time T-Rex has analyzed a stale checkout (same pattern on #396's 1/5 round); if there's a checkout-refresh knob on your side, it may be worth a look.

pHequals7:
The CI `test (meetings)` failure on 4c2ad181 was a pre-existing test race, not this diff: `activeFailureRebuildsSameRoute` asserted `failed.stopCalls == 1` synchronously, but the retired child's stop is dispatched on the async cleanup queue after promotion — under CI load the assertion lands first. Fixed test-only in a4883c02 by waiting for the retirement (matches the test's existing waitUntil pattern). Suite passes locally 3× including full runs; the stale-tree Greptile round's rebuttal is in the comment above.

pHequals7:
Latest P1 (paused failure consumes recovery attempt) — fixed in 3a132bad, and it's the same back-pressure lesson from the mic coordinator, now ported: capture-failure events during pause are ignored entirely (no episode, no dispatch), and a rejected rebuild request refunds the attempt while keeping its cooldown timestamp — so a pause mid-failure can't burn the budget or delay post-resume recovery by a cooldown window. Tests: `captureFailureWhilePausedIsIgnored`, `rejectedRecoveryKeepsCooldown`. Full suite 1674/1674 green locally.

PR #448, opened 2026-08-20
Isolate the config store in tests so the suite stops rewriting the real config

esteugene:
## Summary

Running `swift test` on a machine where Muesli is installed rewrites the user's own settings.

`MuesliController.init` already takes a `configStore:` and `ConfigStore.init` already takes a
`supportDirectory:`, so the seams exist. Several tests just construct them with the defaults,
which resolve to the live support directory.

The clearest case is `MeetingHookIntegrationTests`:

```swift
controller.updateConfig {
    $0.meetingHookEnabled = true
    $0.meetingHookPath = "/definitely/missing/hook.sh"
    $0.meetingHookTimeoutSeconds = 1
}
```

`updateConfig` calls `configStore.save(config)`, and that store points at
`~/Library/Application Support/Muesli/config.json`. The suite passes, and afterwards the
installed app is configured with a hook path that does not exist and a 1 second hook timeout,
with no record of what was there before.

This is not hypothetical. On my machine it silently replaced a working post-meeting hook and
flipped `meeting_recording_save_policy` from `always` to `prompt`. The next real meeting ran
with those settings: the hook never fired, and the recording was not kept, so the audio is gone.
The only trace was one line in `meeting-hook.log` saying the executable does not exist.

It also cannot be worked around from the outside. macOS resolves Application Support from the
user record rather than `$HOME`, so running the suite with `HOME` pointed at a scratch directory
still writes the real file. I verified that directly before writing this.

## Changes

`MeetingsNavigationTests` already has the right shape, injecting
`ConfigStore(supportDirectory: makeSupportDirectory())`. This applies the same idiom to the
remaining places rather than introducing a new convention.

- `MeetingHookIntegrationTests`, `GoogleCalendarTests` and one direct controller construction in
  `MeetingsNavigationTests` now inject a throwaway support directory.
- `ConfigStoreTests` uses a throwaway store for the load, round-trip and permission tests. The
  round-trip no longer has to save the user's config back afterwards, because it never touches
  it. `config path is in Application Support` deliberately keeps the default store: it asserts
  the real location and only reads it.
- The two `FloatingIndicatorController` tests in `NemotronStreamingTests` no longer load the live
  config, so they stop depending on how the machine happens to be set up.

No production code changes.

## Validation

- `swift test --filter "ConfigStoreTests|MeetingHookIntegrationTests|GoogleCalendarTests|MeetingsNavigationTests"`,
  101 tests in 5 suites, all passing.
- Diffed `config.json` before and after a full run of those suites: identical, no keys changed.
- Reverted the patch and re-ran `MeetingHookIntegrationTests`: `meeting_hook_path` became
  `/definitely/missing/hook.sh` and `meeting_hook_timeout_seconds` became `1` again, which is how
  the original breakage was reproduced.
- `scripts/test_ci_test_shards.sh` passes.

## Contribution certification

      under Muesli's
      and licenses or terms.
      resulting changes.

### Third-party materials

None.

### AI assistance

Written with Claude Code. I reviewed the diff, ran the suites, and reproduced both the failure
and the fix on a dev build before opening this.

---
<a href="https://app.blacksmith.sh/Muesli-HQ/codesmith/muesli/pr/448"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789809669&installation_model_id=422865&pr_number=448&repository=Muesli-HQ%2Fmuesli&return_to=https%3A%2F%2Fgithub.com%2FMuesli-HQ%2Fmuesli%2Fpull%2F448&signature=be720fb337d84c88501a78c506c8e6b82d31740f4ed8bd4a4da37efea2ff7fde"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **Tests**
  - Improved test isolation by using temporary configuration directories.
  - Reduced interference between tests involving configuration, calendar placeholders, meetings, and warning displays.
  - Updated persistence and permission checks to avoid affecting default application settings.

### DIFF
--- a/native/MuesliNative/Tests/MuesliTests/ConfigStoreTests.swift
+++ b/native/MuesliNative/Tests/MuesliTests/ConfigStoreTests.swift
@@ -6,18 +6,28 @@ import MuesliCore
 @Suite("ConfigStore", .serialized)
 struct ConfigStoreTests {
 
+    /// A throwaway support directory. Tests must never read or write the real
+    /// config.json: `swift test` runs on machines where Muesli is installed, and
+    /// a test that saves into the live support directory rewrites the settings of
+    /// whoever is running the suite.
+    private func makeStore() -> ConfigStore {
+        ConfigStore(
+            supportDirectory: FileManager.default.temporaryDirectory
+                .appendingPathComponent("muesli-config-store-\(UUID().uuidString)", isDirectory: true)
+        )
+    }
+
     @Test("load returns a valid config")
     func loadReturnsConfig() {
-        let store = ConfigStore()
+        let store = makeStore()
         let config = store.load()
-        // Hotkey may have been customized by user — just verify it loaded
         #expect(HotkeyConfig.label(for: config.dictationHotkey.keyCode) != nil)
         #expect(!config.sttBackend.isEmpty)
     }
 
     @Test("save and load round-trip")
     func saveLoadRoundTrip() {
-        let store = ConfigStore()
+        let store = makeStore()
         let original = store.load()
 
         var config = original
@@ -40,9 +50,6 @@ struct ConfigStoreTests {
         #expect(loaded.whisperLanguage == WhisperKitLanguage.german.rawValue)
         #expect(loaded.appleSpeechLanguage == "en-US")
         #expect(loaded.meetingSummaryBackend == "openrouter")
-
-        // Restore original
-        store.save(original)
     }
 
     @Test("config path is in Application Support")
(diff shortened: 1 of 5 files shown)


esteugene:
Closing this: it is a contributor-only papercut rather than something that affects the app, and I would rather not spend maintainer review time on it. Happy to reopen if you ever want the test isolation tidied up.

